PR #154, opened 2019-07-08
Message output now shows the delimiter used

### LINKED ISSUE #95: Add guessed delimiter to message output
We never say what delimiter has been guessed currently, which we probably should.

R3myG:
Fixes #95


jimhester:

Thanks!! You are da bomb!

Issue #145, opened 2019-06-26
Throw a better error message when trying to write to a zip file, which is not supported


jimhester:
```r
vroom_write(mtcars, "test.zip")
#> Error in utils::unzip(path, list = TRUE) :
#>   zip file 'test.zip' cannot be opened
```

PR #155, opened 2019-07-08
Thow error message when trying to write a zip file

meta00:
fix issue #145

### DIFF
--- a/R/path.R
+++ b/R/path.R
@@ -47,21 +47,22 @@ standardise_one_path <- function (path, write = FALSE) {
     )
   }
 
+  ext <- tolower(tools::file_ext(path))
+
   if (!write) {
     path <- check_path(path)
   } else {
     path <- normalizePath(path, mustWork = FALSE)
+    if (ext == "zip") {
+      stop("Can only read from, not write to, .zip", call. = FALSE)
+    }
   }
 
-  switch(tolower(tools::file_ext(path)),
+  switch(ext,
     gz = gzfile(path, ""),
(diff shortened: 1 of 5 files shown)


R/path.R
```diff
@@ -57,7 +57,11 @@ standardise_one_path <- function (path, check = TRUE) {
     gz = gzfile(path, ""),
     bz2 = bzfile(path, ""),
     xz = xzfile(path, ""),
-    zip = zipfile(path, ""),
+    zip = if(check){
+            zipfile(path, "")
+          } else {
+            stop("Writing zip files is not supported")
```
hadley: I think it would be a little cleaner to do the check in the `if (check)` block above.  But I think the `check` argument name is confusing, and it would be more informative if it was called `write`. Would you mind making that change too?

meta00: I think `check` should be called `read` instead of `write` no?

hadley: You'd have to flip the value, but I think `write` would make it more clear to me

hadley: You'll also need to update the block that use `check` above, and that's where I'd recommend you also perform this check.

meta00: I did in the commit https://github.com/r-lib/vroom/pull/155/commits/16db47d51130a70345a9d0f0ec4ac54cb41e2b3c

hadley: Oh sorry, somehow github didn't show that to me.

hadley:
Thanks @meta00!

Issue #125, opened 2019-05-31
feature request: handling multiple files in a zip


randomgambit:
Hi @jimhester I am trying to use `vroom` on `zip` archives that contain multiple files.

At the moment, I think `vroom` reads the first file in the zip by default. Indeed, I created a `zip` archive containing two copies of the example files we used yesterday:

```
> test <- vroom('/mypath/myzipped.zip', col_names = c('a','b','c','d'))
Multiple files in zip: reading 'file.out'
Observations: 1,167                                                                                                                
Variables: 4
chr [3]: a, b, c
dbl [1]: d
```
Only the first file `file.out` is being read. I wonder if `vroom` could directly read all the files in the archive in one go. What do you think?

jimhester:
I don't think we could do this automatically, too many zip files have heterogeneous files. You can write a wrapper function to do this without a lot of work however, e.g.

```r
read_all_zip <- function(file, ...) {
  filenames <- unzip(file, list = TRUE)$Name
  vroom(purrr::map(filenames, ~ unz(file, .x)), ...)
}
```

randomgambit:
thanks this is what I ended up doing but I wondered if using `unz` every time was inefficient.

randomgambit:
@jimhester yes, good idea. Something unrelated: I get 150MB/s with vroom using connections. I wonder if I should use `VROOM_CONNECTION_SIZE` to improve the speed. Would that help?

jimhester:
No, not likely.

randomgambit:
makes sense  thanks for the clarification

PR #156, opened 2019-07-08
add example to handle multiple files in a zip

meta00:
fix issue #125

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,5 @@
 # vroom (development version)
 
-* Add a section to getting started vignette to show to how handle multiple files in a zip (@meta00, #125) 
-
 # vroom 1.0.2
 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* Add a section to getting started vignette to show to how handle multiple files in a zip (@meta00, #125)
```
hadley: You don't need a news bullet here, since it's just a documentation change. Can you please remove it?

meta00: done in commit https://github.com/r-lib/vroom/pull/156/commits/06c8cbfd2c388939ac61395ecc5e1ad0b2a2a2d3

hadley:
Thanks @meta00!

PR #157, opened 2019-07-08
Documentation about the reading of compressed files.

### LINKED ISSUE #121: Document how reading compressed files work
There has been more than one person wondering about how vroom works with compressed files, if they are uncompressed to a temporary directory etc.

We should add information to http://vroom.r-lib.org/articles/vroom.html#reading-compressed-files explaining how this works.

vroom compresses the file during indexing and writes the uncompressed file to a temporary file in the `VROOM_TEMP_PATH` (defaults to `tempdir()`).

The temporary file remains until it is no longer needed, e.g. the dataset is removed (and garbage collected), all of the data has been materialized or the R session ends. Once this happens the temporary file is automatically cleaned up.

CriscelyLP:
Including changes in the documentation about compressed files. Fixed #121


vignettes/vroom.Rmd
```diff
@@ -112,6 +112,12 @@ file <- vroom_example("mtcars.csv.gz")
 vroom(file)
 ```
 
+About how it work `vroom()`. After the works of this function (indexing and writes) the already uncompressed file is going to compressed again but into a temporal file (in a `tempdir()`).
```
batpigandme: I think you may mean temporary where you're using temporal?

batpigandme: I think you can collapse the first fragment with the second full-sentence (see suggestion)

batpigandme:
```suggestion
After the `vroom()` function indexes and writes the initial file, the already uncompressed file is going to compressed again but into a temporary file (in a `tempdir()`).
```

vignettes/vroom.Rmd
```diff
@@ -112,6 +112,12 @@ file <- vroom_example("mtcars.csv.gz")
...
+
+The temporal file created remains until it is no longer needed, e.g. when the dataset is removed and all of the data has been materialiwed or the R session ends. After this the temporal file is automatically cleaned up.
```
batpigandme:
```suggestion
The temporary file created remains until it is no longer needed, e.g. when the dataset is removed and all of the data has been materialised, or when the R session ends. After this, the temporary file is automatically cleaned up. 
```

vignettes/vroom.Rmd
```diff
@@ -112,6 +112,12 @@ file <- vroom_example("mtcars.csv.gz")
 vroom(file)
 ```
 
+About how it work `vroom()`. After the works of this function (indexing and writes) the already uncompressed file is going to compressed again but into a temporal file (in a `tempdir()`).
+
+The temporal file created remains until it is no longer needed, e.g. when the dataset is removed and all of the data has been materialiwed or the R session ends. After this the temporal file is automatically cleaned up. 
+
+The temporal file is not going to be saved by default. In case you would like save this temporal file for future uses, see `vroom_write`.
```
batpigandme:
```suggestion
The temporary file is not saved by default. To save this temporary file for future uses, see `vroom_write()`.
```

batpigandme: Note: We follow function names with parens! 👍

vignettes/vroom.Rmd
batpigandme: I think this might be missing an initial pass of decompression (cc @jimhester ?)

CriscelyLP: What you mean with that?

CriscelyLP:
Fixes #121

vignettes/vroom.Rmd
CriscelyLP: Yes, thank you! :)

batpigandme: "About how it work `vroom()`" isn't actually a full sentence. I think it's implied from the surrounding context and language that you're talking about how the function works, so in my suggestion, I brought the mention of `vroom()` into that first full sentence (second, if you're counting the initial frag!)

batpigandme: So, I think this is resolved, since you accepted those edits!

Shians:
I'm curious about this, the proposed changes are a bit confusing. 

> After the `vroom()` function indexes and writes the initial file, the already uncompressed file is going to compressed again but into a temporary file (in a `tempdir()`).

I'm not sure what is meant by "indexes and writes the initial file", is there some write event that occurs when you read a file with vroom?

It would be nice to have some motivation for this strange sequence of events. It sounds like a file is decompressed, indexed, and a duplicate compressed file is created. Is the newly created compressed file identical to the original compressed file? If not, how is the new file organised that assists vroom?

Thanks,
Shian.

PR #159, opened 2019-07-08
add name repair section to vignette

meta00:
fix issue #81

### DIFF
--- a/vignettes/vroom.Rmd
+++ b/vignettes/vroom.Rmd
@@ -250,9 +250,11 @@ vroom(
 
 Often the names of columns in the original dataset are not ideal to work with. `vroom()` uses the same .name_repair argument as tibble, so you can use one of the default name repair strategies or provide a custom function A great approach is to use the [`janitor::make_clean_names()`](http://sfirke.github.io/janitor/reference/make_clean_names.html) function as the input.
 
-```{r}
-vroom(vroom_example("mtcars.csv"),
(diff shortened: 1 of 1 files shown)


vignettes/vroom.Rmd
```diff
@@ -246,6 +246,16 @@ vroom(
 )
 ```
 
+## Name repair
+
+Often the names of columns in the original dataset are not ideal to work with. `vroom()` uses the same .name_repair argument as tibble, so you can use one of the default name repair strategies or provide a custom function A great approach is to use the [janitor] (http://sfirke.github.io/janitor/) `make_clean_names()` function as the input.
```
hadley:
```suggestion
Often the names of columns in the original dataset are not ideal to work with. `vroom()` uses the same `.name_repair` argument as tibble, so you can use one of the default name repair strategies or provide a custom function. Another great approach is to use the [janitor](http://sfirke.github.io/janitor/) `make_clean_names()` function as the input.
```

hadley: But maybe it would be to link `janitor::make_clean_names()` directly to its reference page?

meta00: you meant like this https://github.com/r-lib/vroom/pull/159/commits/fb71963bbd02cdd4cd9947599d1e113345781b5b ?

hadley:
Thanks @meta00!

PR #160, opened 2019-07-08
allow %d and %m to take either 1 or 2 characters, e.g. 03/04/1999 or 3/4/1999

edzer:
The problem seems that dates like 1/5/1999 are not read with
format "%m/%d/%Y", because 01/05/1999 is expected. I think
that if you read ?strptime well, this is correct, but then base
R works differently and accepts single digit days and months. So,
I think the prefered behavior is to accept them.

This PR is messy and full with print statements. I believe that the
line marked with "need to set back the parser iterator/pointer one
character here?" should be replaced with setting back the parser
pointer one step, as we have a fail read here which does NOT raise
an error. The error then happens in next step, when the "/" (or other
separator) has been absorbed, I suspect. Happy to revisit this when
I know how the parser works. Please ignore otherwise.


edzer:
I think I fixed it, now.

edzer:
Hi Jim, any news on this one?

jimhester:
Sorry, I did not see that you had resolved this and had been working on other packages since July, thank you for pinging me, if you could 
please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`, then I can merge this. Thank you for working on it!

jimhester:
Thanks for working on this Edzer!

Thanks!! You are da bomb!

PR #252, opened 2020-07-28
Typo? Unless sec^2 was intentional

lwjohnst86:
Fixed typo in stating speed of vroom. I assume it isn't Gb/sec/sec? Unless that's it's acceleration, then well this can be ignored!


jimhester:
Thanks, not sure how that happened...

Issue #313, opened 2021-02-18
Reading data with decimal mark ',' and col_type 'd' gives incorrect values


kent37:
Reading data that uses comma as the decimal separator and specifying double format doesn't work - the values read as NA. It reads correctly if specified as 'number' format.

``` r
library(vroom)

# Data using comma as the decimal mark
df = tibble::tribble(
  ~value,
  "23,4"
)

# Make a test data file
temp_path = tempfile()
vroom_write(df, temp_path)

# Reading with decimal mark ',' and col_type 'd' fails
df = vroom(temp_path, delim='\t',
           locale=locale(decimal_mark=','),
           col_types=cols('d'))

print(df)
#> Warning: One or more parsing issues, see `problems()` for details
#> # A tibble: 1 x 1
#>   value
#>   <dbl>
#> 1    NA

problems(df)
#> # A tibble: 1 x 5
#>     row   col expected actual file                                              
#>   <int> <int> <chr>    <chr>  <chr>                                             
#> 1     2     1 a double 23,4   C:/Users/kjohnson/AppData/Local/Temp/Rtmpi0fHjm/f~

# Reading with decimal mark ',' and col_type 'n' gives correct result
df = vroom(temp_path, delim='\t',
           locale=locale(decimal_mark=','),
           col_types=cols('n'))

print(df)
#> # A tibble: 1 x 1
#>   value
#>   <dbl>
#> 1  23.4

problems(df)
#> # A tibble: 0 x 5
#> # ... with 5 variables: row <int>, col <int>, expected <chr>, actual <chr>,
#> #   file <chr>
```

<sup>Created on 2021-02-18 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

jimhester:
Fixed by https://github.com/r-lib/vroom/pull/319

PR #319, opened 2021-03-03
Correctly read double values with comma as decimal separator #313

kent37:
Fixed by providing `decimalMark` to `bsd_strtod` instead of using hard-coded `'.'`.


jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jimhester:
Thank you!

PR #321, opened 2021-04-08
Use the tz and date headers from clock rather than localtime.c

jimhester:
@DavisVaughan How do you envision packages using the tz library from clock? Because the code I use here uses `local_days`, which is defined in the non-header only tz I need to include `tz.cpp` in my code.

This works 'as-is' on my local macOS machine (and likely linux) using the OS supplied timezone databases, however it won't work on windows.

Are packages supposed to use your tzdb package and then set the timezone database to that in their `.onLoad()` or is there another way to go about this? Ideally without calling into R functions in clock from C++?


jimhester:
Closed in favor of #322

PR #325, opened 2021-04-19
Replace usage of clock with tzdb

DavisVaughan:
This PR replaces the usage of clock with [zones](https://github.com/r-lib/zones).

zones is a lightweight way to expose the date library's header API, along with 3 callables for working with time zones. As of [this PR](https://github.com/r-lib/clock/pull/222), clock now uses zones as well, rather than embedding date directly.

I am hopeful that compilation times for vroom should be much faster now. For example, on vroom's Ubuntu r-devel check I see:

> ✔ Built zones 0.0.0.9000 (15.2s)

VS this with clock

> ✔ Built clock 0.2.0 (2m 38.3s)

At the advice of other r-lib co-workers on Slack, the callables now require a `std::string& error` argument. If this is not empty after calling a zones callable, that is the signal that an error has occurred when either loading the time zone or looking up the local time. This is safer than throwing exceptions across packages, at the cost of putting a little more burden on the client (here, vroom).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,10 +46,10 @@ Imports:
     rlang (>= 0.4.2),
     stats,
     tibble (>= 2.0.0),
+    tzdb (>= 0.1.0.9001),
     vctrs (>= 0.2.0),
     tidyselect,
-    withr,
-    zones (>= 0.0.0.9000)
+    withr
 Suggests: 
     bench (>= 1.1.0),
     covr,
@@ -73,7 +73,7 @@ Suggests:
 LinkingTo: 
     progress (>= 1.2.1),
     cpp11 (>= 0.2.0),
-    zones (>= 0.0.0.9000)
+    tzdb (>= 0.1.0.9001)
 VignetteBuilder: 
     knitr
 Config/testthat/edition: 3
@@ -86,4 +86,4 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1
 SystemRequirements: C++11
 Remotes:
-    r-lib/zones
+    r-lib/tzdb
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,7 +39,6 @@ Imports:
     bit64,
     crayon,
     cli,
-    clock (>= 0.2.0),
     glue,
     hms,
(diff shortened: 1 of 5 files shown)


jimhester:
Rather than returning an error as a string I think a more idiomatic API would be to return a [error_condition](https://en.cppreference.com/w/cpp/error/error_condition)?

Also rather than introducing another new package, maybe we could convert tzdb to be what is in zones?

DavisVaughan:
@jimhester I've just pushed an example implementation that uses `error_condition`s. They are a bit tricky!

Unfortunately, we lose two major things here:

First, I cannot figure out a way to surface the `.what()` message of date's `runtime_error`s up into the error condition objects. So it seems like we are going to lose that. 99% of the time, the only runtime error would be the one that could occur when you look up an invalid time zone, so this isn't that big of a deal I guess.

Second, it normally seems like you check an `error_condition` object against the relevant error code directly with `==` to determine if it matches or not. Like: `error_condition cnd = locate_zone(); if (cnd == zones::zones_errc::success) "yay"`. However, the underlying `error_category` object is a singleton, and it is used as part of the equivalence check. From what I can tell, the condition object returned by `locate_zone()` (created on the zones side) has a different `error_category` singleton than the one that gets made from the error code you check against (created on the vroom side), so the `==` always returns `false`. To get around this, I exposed `zones::is_error(std::error_condition)` instead, which ensures the singleton in zones is consistently used for both sides of the `==` operation.

Since we lose information, and since the singleton bit is a little complicated, I think it might just be simpler to return `true` in the case of success, and `false` in the case of failure from the zones API functions. Do you have any issue with that?

Here is the implementation on the zones side, if you are interested in taking a look / have any advice https://github.com/r-lib/zones/pull/1

DavisVaughan:
This seems unrelated, but strange? https://github.com/r-lib/vroom/pull/325/checks?check_run_id=2383448940#step:10:24

jimhester:
Sure, if the only states are success or failure just returning true or false seems fine.

DavisVaughan:
@jimhester I've merged zones into tzdb!

jimhester:
Thanks a ton!

PR #358, opened 2021-07-19
All missing -> assume character

zeehio:
Inspired by https://twitter.com/cboettig/status/1417177718949498888?s=20

I can fix failing tests and add a news entry to this pull request of you think the change is worth it.

### DIFF
--- a/src/guess_type.cc
+++ b/src/guess_type.cc
@@ -113,7 +113,7 @@ std::string guess_type__(
   }
 
   if (allMissing(input)) {
-    return "logical";
+    return "character";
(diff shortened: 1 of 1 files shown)


jennybc:
This matter (the type of a column with only `NA`s) is something that's been discussed elsewhere, at length (most recently, yesterday, in https://github.com/tidyverse/dplyr/issues/6097). The current position throughout the tidyverse is that logical is the best choice. To be sure, this has advantages and disadvantages, but at this point we think it's a net positive overall.

PR #359, opened 2021-07-21
Hack return value of compare_proxy

hadley:
I think this is the simplest fix - the R CMD check of method/generic consistency isn't a problem because the method is registered dynamically. If you think its important, I could do the comparison once, on load, but it doesn't seem that important to me.


jimhester:
Thanks a watt!

PR #386, opened 2021-11-23
Avoid an uninitialized `bool` for the CRAN UBSAN check

DavisVaughan:
i.e. fixes:

```cpp
/usr/local/bin/../include/c++/v1/__utility/pair.h:192:52: runtime error: load of value 128, which is not a valid value for type 'bool'
    #0 0x7f1daea6b953 in std::__1::pair<bool, bool>::pair<bool, bool&, false>(bool&&, bool&) /usr/local/bin/../include/c++/v1/__utility/pair.h:192:52
    #1 0x7f1daea6b3f9 in vroom::is_blank_or_comment_line(char const*, char const*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool) /data/gannet/ripley/R/packages/tests-clang-SAN/vroom/src/./utils.h:68:12
```

The first commit is the minimal change required to fix this. I just avoided creating `has_comment` entirely, since it didn't seem to have anything to do with the first half of the if statement.

The second commit slightly refactors the function in a way that I think makes it easier to follow. I originally had trouble knowing what the return value meant, and had to go look for locations where `is_blank_or_comment_line()` was being used to understand how the two `bool` values should be interpreted. Making them local variables with names makes it a bit clearer, IMO. I can revert this one if you disagree.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* Fix uninitialized bool detected by CRAN's USBAN check (https://github.com/r-lib/vroom/pull/386)
+
 * Fix buffer overflow when trying to parse an integer field that is over 64 characters long (https://github.com/tidyverse/readr/issues/1326)
 
 * Fix subset indexing when indexes span a file boundary multiple times (#383)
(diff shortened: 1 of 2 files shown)


jimhester (approved):

jimhester:
Looks good to me and I like the refactoring as well.

PR #390, opened 2021-12-01
Fix recursive gc error that could occur with POSIXct columns

### LINKED ISSUE #389: Recursive gc error
```R
library(dplyr)
library(nycflights13)
library(magrittr)

path <- tempfile()
dir.create(path)

gctorture(TRUE)

flights %>% 
  group_by(month) %>% 
  group_walk(~ vroom::vroom_write(.x, sprintf("%s/month-%02i.csv", path, .y$month)))
```

```
*** recursive gc invocation
*** recursive gc invocation
Error in vroom_write_(xx, file, delim, eol, na_str = na, col_names = col_names,  :
  cannot get ALTSTRING_ELT during GC
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
sive gc invo>c aion
** ecrsvegcinoctin
** ecursive gc invocation
*** recursive gc invocation
*** recursive gc invocation
*** recursive gc invocation
*** recursive gc invocation
*** recursive gc invocation
*** recursive gc invocation
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
```

From https://github.com/tidyverse/multidplyr/issues/127#issuecomment-983449321

DavisVaughan:
Closes #389 

The analysis in this comment was correct https://github.com/r-lib/vroom/issues/389#issuecomment-983667565, the `output_column()` method for POSIXct was creating another character ALTREP vector _after_ `vroom_convert()` was called.

I didn't add a test, because this is hard to test for, but I confirmed that the reprex here no longer errors https://github.com/r-lib/vroom/issues/389#issuecomment-983691057

---

POSIXct isn't necessarily the only type where this could happen.

`output_column.POSIXt()` calls `format()`, and what is happening in `format()` is:

* There is some C code that converts POSIXct -> character and returns a non-altrep character vector called `y`
* At the R level, there is some code that adds names onto `y`
* `names<-` at the C level will create an “ALTREP wrapper” around `y`. It allows you to wrap `y` without duplicating it, and then add new attributes on top of it (like names) cheaply

```r
# part of format.POSIXlt
y <- .Internal(format.POSIXlt(x, format, usetz))
names(y) <- names(x$year)
```

* There is additionally a `structure()` call in `format.POSIXct()` that tries to add names on _again_ to the result of `format.POSIXlt()`. This is probably also creating another ALTREP wrapper layer on top of the first one.

```r
# part of format.POSIXct
structure(format.POSIXlt(as.POSIXlt(x, tz), format, usetz,  ...), names = names(x))
```

So it is entirely possible that other `output_column()` methods could do this, but I don’t think any of the currently implemented ones do.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* Fix recursive gc error that could occur when writing POSIXct columns (https://github.com/r-lib/vroom/issues/389)
+
 # vroom 1.5.7
 
 * Jenny Bryan is now the official maintainer.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # vroom (development version)
 
-* Fix recursive gc error that could occur when writing POSIXct columns (https://github.com/r-lib/vroom/issues/389)
+* Fix recursive gc error that could occur during `vroom_write()` when `output_column()` generates an ALTREP vector (https://github.com/r-lib/vroom/issues/389)
 
 # vroom 1.5.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # vroom (development version)
 
-* Fix recursive gc error that could occur during `vroom_write()` when `output_column()` generates an ALTREP vector (https://github.com/r-lib/vroom/issues/389)
+* Fix recursive gc error that could occur during `vroom_write()` when `output_column()` generates an ALTREP vector (#389).
 
 # vroom 1.5.7
 
--- a/R/vroom_write.R
+++ b/R/vroom_write.R
@@ -46,6 +46,7 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names 
     )
   }
 
+  input <- x
 
   quote <- match.arg(quote)
   escape <- match.arg(escape)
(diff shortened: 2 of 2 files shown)


romainfrancois (approved):
LGTM

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* Fix recursive gc error that could occur during `vroom_write()` when `output_column()` generates an ALTREP vector (https://github.com/r-lib/vroom/issues/389)
```
jennybc, line 3:
```suggestion
* Fix recursive gc error that could occur during `vroom_write()` when `output_column()` generates an ALTREP vector (#389).
```

jennybc (approved):
It's too bad we can't add a test, but I agree that it's impractical.

I've tried to look at this with a Chesterton's Fence perspective and it seems likely that the initial order (materialize, convert to character) seemed most natural and it's easy to overlook that character conversion can re-introduce ALTREP vectors (especially since POSIXt may be the only current example).

PR #399, opened 2022-02-04
Allow non ascii filenames

### LINKED ISSUE #394: File path encoding problem on Windows
Manual transfer of https://github.com/tidyverse/readr/issues/1345

``` r
library(vroom)

vroom("C:/Users/jenny/Downloads/Renda Fixa Pré.csv")
#> Rows: 0 Columns: 0
#> 
#> i Use `spec()` to retrieve the full column specification for this data.
#> i Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> # A tibble: 0 x 0

# this should be happening automatically
vroom(
  iconv(
    "C:/Users/jenny/Downloads/Renda Fixa Pré.csv",
    to = "UTF-8"
  )
)
#> Rows: 15 Columns: 5
#> -- Column specification --------------------------------------------------------
#> Delimiter: ","
#> chr (2): Nome, Tipo
#> dbl (2): Prazo, InvestimentoInicial
#> 
#> i Use `spec()` to retrieve the full column specification for this data.
#> i Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> # A tibble: 15 x 5
#>    Nome                                       Taxa Tipo  Prazo InvestimentoInic~
#>    <chr>                                     <dbl> <chr> <dbl>             <dbl>
#>  1 CDB Caruana Pre                              22 PRÉ     365              1000
#>  2 CDB Fator Pré-fixado                         22 PRÉ     365              5000
#>  3 CDB NBC Pré-fixado                           22 PRÉ     365              1000
#>  4 CDB BDMG Pré                                229 PRÉ     365             10000
#>  5 CDB BRPartners Pre                          233 PRÉ     365             20000
#>  6 CDB BCG Brasil Pré-fixado                   251 PRÉ     365            100000
#>  7 CDB Pine Pré-fixado                         258 PRÉ     365              5000
#>  8 CDB Modal Pré-fixado                        279 PRÉ     365              1000
#>  9 CDB MAXINVEST-RNX PRE                        29 PRÉ     365              1000
#> 10 CDB Agibank Pré-Fixado                      301 PRÉ     365              1000
#> 11 CDB Banco Industrial do Brasil Pré-fixado   309 PRÉ     365              5000
#> 12 CDB Luso Pre                                 31 PRÉ     365              5000
#> 13 CDB Omni Pré-Fixado                          32 PRÉ     365              5000
#> 14 CDB Avista Pre                              369 PRÉ     365              1000
#> 15 LC Avista Pre                               369 PRÉ     365              1000
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

sbearrows:
Closes #394 

I'm using the CI to check that my test fails on Windows before I make the fix (since I don't have a Windows machine yet)


tests/testthat/test-path.R
```diff
@@ -95,3 +95,16 @@ test_that("can write to a tar.gz file if the archive package is available", {
...
 })
+
+test_that("can read filepaths with non-ASCII characters", {
+  skip_on_cran()
+  ## chosen to be non-ASCII but
+  ## [1] representable in Windows-1252 and
+  ## [2] not any of the few differences between Windows-1252 and ISO-8859-1
+  ## a-grave + e-diaeresis  + Eth + '.csv'
+  tricky_filename <- "\u00C0\u00CB\u00D0\u002E\u0063\u0073\u0076"
+  file.copy(vroom_example("mtcars.csv"), tricky_filename)
```
jennybc, line 106: On principle, in terms of the unit test mindset, I suggest using the very smallest example file or even writing the file yourself. It can have 1 variable and 1 row. It just needs to exist. The only reason the readxl example uses `file.copy()` on a built-in example is because we can inline file creation `.xls` or `.xlsx`. But it's very easy to do create a small file that `vroom()` to read. You might find a good example of this pattern elsewhere in the vroom or readr tests (I haven't looked).

tests/testthat/test-path.R
```diff
@@ -95,3 +95,16 @@ test_that("can write to a tar.gz file if the archive package is available", {
...
+  expect_true(file.exists(tricky_filename))
+  on.exit(file.remove(tricky_filename))
```
jennybc, line 108: On principle, even though it does not matter here, I would add `add = TRUE` to the `on.exit()` call and place it right after the file creation code. This is always safest. The `add = TRUE` is important because it really should be the default for `on.exit()`: usually you want to add stuff to do later, not replace all previously scheduled todo's (note that readxl's tests currently have some mistakes in this regard!). The placement-right-after-creation is to minimize the chance that something else goes wrong, before the `on.exit()` call, resulting in us failing to clean up after ourselves. All of this is general advice about scheduling actions to happen later.

But here in this test, I'd actually prefer that you use `withr::with_tempfile()` or, probably, `withr::local_tempfile()`. In general, I'd prefer that we convert vroom's tests to use withr, whenever we touch them.

https://withr.r-lib.org/reference/with_tempfile.html

These are both nitpicky things. I'm just using this opportunity to discuss this topics, which will come up often in the future.

tests/testthat/test-path.R
```diff
@@ -95,3 +95,16 @@ test_that("can write to a tar.gz file if the archive package is available", {
...
+  expect_snapshot(vroom(tricky_filename))
```
jennybc, line 109: I suggest replacing this with an `expect_equal()` for the return value of `vroom()`. Usually we save snapshot tests for testing printed output that's really hard to test otherwise. Once you change the test file to a small built-in example or to a small file you create in this test, it should be easy to convert this to `expect_equal()`.

sbearrows: Alright, thanks for the feedback! I was about to add another test for multiple files because I think both `standardise_one_path()` and `standardise_path()` will need `enc2utf8(path)`.
I'll make these edits first.

PR #403, opened 2022-02-08
Add Kevin's test

jennybc:
I removed the non-ascii text from a comment.

Relates to #394, #399 (Shelby's PR), #403 (Kevin's PR, from which I just took the test).

### DIFF
--- a/tests/testthat/test-path.R
+++ b/tests/testthat/test-path.R
@@ -95,3 +95,28 @@ test_that("can write to a tar.gz file if the archive package is available", {
   expect_equal(res$path, "mtcars")
   expect_equal(res$size, 1281)
 })
+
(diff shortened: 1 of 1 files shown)


jennybc:
I'm beginning to wonder if this whole thing is actually the RStudio problem?

https://github.com/rstudio/rstudio/issues/10507

I definitely repro'd on my Windows VM *from within RStudio*.
And none of us have ever gotten the new test to fail on the Windows jobs in the build matrix, even before introducing the fix.

cc @kevinushey @sbearrows 

I think the fix may be a good idea regardless, but I would like to better understand what's actually going on.

jennybc:
It's not the RStudio problem. It's an unexpected interaction between the encoding of the file path and presence (absence, in the case of the OP's file) of a trailing newline.

PR #408, opened 2022-03-03
'usethis::use_tidy_description()'

sbearrows:
Part of #407

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,37 +1,23 @@
 Package: vroom
 Title: Read and Write Rectangular Text Data Quickly
 Version: 1.5.7.9000
-Authors@R: 
-    c(person(given = "Jim",
-             family = "Hester",
-             role = "aut",
-             email = "jim.hester@rstudio.com",
-             comment = c(ORCID = "0000-0002-2739-7082")),
-      person(given = "Hadley",
-             family = "Wickham",
-             role = "aut",
-             email = "hadley@rstudio.com",
-             comment = c(ORCID = "0000-0003-4757-117X")),
-      person(given = "Jennifer",
-             family = "Bryan",
-             role = c("aut", "cre"),
-             email = "jenny@rstudio.com",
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -78,16 +64,14 @@ Suggests:
...
+    progress (>= 1.2.1),
     tzdb (>= 0.1.1)
 VignetteBuilder: 
     knitr
+Config/Needs/website:nycflights13, tidyverse/tidytemplate
 Config/testthat/edition: 3
 Config/testthat/parallel: false
-Config/Needs/website:
-    nycflights13,
-    tidyverse/tidytemplate
```
jennybc, line 90: I would accept the new **location** of `Config/Needs/website`, but not the re-formatting. This will eventually get fixed in desc, but in the meantime, just retain the preferred formatting "by hand".

For background:
https://github.com/r-lib/desc/issues/117
https://github.com/r-lib/desc/issues/122

DESCRIPTION
```diff
@@ -78,16 +64,16 @@ Suggests:
...
     cpp11 (>= 0.2.0),
+    progress (>= 1.2.1),
     tzdb (>= 0.1.1)
 VignetteBuilder: 
     knitr
-Config/testthat/edition: 3
-Config/testthat/parallel: false
 Config/Needs/website:
-    nycflights13,
+    nycflights13,
```
jennybc, line 73: This is very much a nitpick. But there should be no trailing whitespace here. It's a good nudge to configure any text editor / IDE you're using to remove trailing whitespace in files treated as source code (but you would not want this for, say, markdown). Good default config will pay off in the long run.

In RStudio, do *Tools > Global Options > Code > Saving* and I recommend:
* YES Ensure that source files end with newline
* YES Strip trailing horizontal whitespace when saving <-- SPECIFIC THING WE'RE TALKING ABOUT RIGHT NOW

jennybc (approved):
LGTM

nikolajsumik5-commits (approved):

PR #409, opened 2022-03-04
`usethis::use_tidy_style()`

sbearrows:
Part of #407

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -50,10 +50,12 @@
 #'
 #' t1 <- cols(
 #'   column_one = col_integer(),
-#'   column_two = col_number())
+#'   column_two = col_number()
+#' )
 #'
 #' t2 <- cols(
-#'  column_three = col_character())
+#'   column_three = col_character()
+#' )
 #'
 #' t3 <- t1
 #' t3$cols <- c(t1$cols, t2$cols)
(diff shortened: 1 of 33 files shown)


R/col_types.R
```diff
@@ -199,12 +202,13 @@ format.col_spec <- function(x, n = Inf, condense = NULL, colour = crayon::has_co
   delim <- x$delim
 
   if (!is.null(delim) && nzchar(delim)) {
-    delim <- paste0('.delim = ', double_quote(delim), '')
+    delim <- paste0(".delim = ", double_quote(delim), "")
```
jennybc, line 205: This look suspicious to me. Given the presence of a call to `double_quote()`, I suspect there could be a very good reason for using single quotes. You need to scrutinize this.

jennybc: Yeah, I think this does cause test failure, BTW.

sbearrows: I did go through each file that changed, line by line, and tried to scrutinize these changes and there were some changes that I ignored. But as you can see there were a lot of files. I've also never come across style changes that would break tests so that was not something that was on my mind but it will be in the future.

jennybc: Oh, I am sure you did look through this all carefully! Sorry if my "you need to scrutinize this" sounded like a general comment. I wrote it when I first started to look at this and thought we'd actually be merging this PR. And I wanted you to revisit this spot in particular. It is actually a case where the tests worked and would have lead to further investigation anyway.

But I now propose you note our decision to walk away from this in #407, strike the `use_tidy_style()` item out, and close this unmarked.

sbearrows: Ah, gotcha! I'll make note of this in #407 and close this PR.

PR #411, opened 2022-03-07
Update GHA

sbearrows:
Part of #407

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -16,7 +16,7 @@ jobs:
   R-CMD-check:
     runs-on: ${{ matrix.config.os }}
 
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }}) ${{ matrix.config.locale }}
 
     strategy:
       fail-fast: false
@@ -35,6 +35,7 @@ jobs:
           - {os: ubuntu-18.04,   r: 'oldrel-2'}
           - {os: ubuntu-18.04,   r: 'oldrel-3'}
           - {os: ubuntu-18.04,   r: 'oldrel-4'}
+          - {os: ubuntu-18.04,   r: 'release', cran: "https://demo.rstudiopm.com/all/__linux__/bionic/latest", locale: 'en_US'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -42,6 +43,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Set locale
+        if: matrix.config.locale == 'en_US'
+        run: |
+          sudo locale-gen en_US
+          echo "LC_ALL=en_US" >> $GITHUB_ENV
 
       - uses: r-lib/actions/setup-pandoc@v2
 
@@ -57,5 +64,3 @@ jobs:
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
-        with:
-          upload-snapshots: true
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -64,3 +64,5 @@ jobs:
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
+        with:
+          upload-snapshots: true
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -16,7 +16,7 @@ jobs:
   R-CMD-check:
     runs-on: ${{ matrix.config.os }}
 
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }}) ${{ matrix.config.locale }}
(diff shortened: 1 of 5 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
   R-CMD-check:
     runs-on: ${{ matrix.config.os }}
 
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }}) ${{ matrix.config.locale }}
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
```
jennybc, line 20: We need to keep `${{ matrix.config.locale }}` here and several of the other changes below. Their purpose is to test vroom in a non-UTF-8 locale, which we should retain.

.github/workflows/R-CMD-check.yaml
```diff
@@ -35,7 +35,6 @@ jobs:
           - {os: ubuntu-18.04,   r: 'oldrel-2'}
           - {os: ubuntu-18.04,   r: 'oldrel-3'}
           - {os: ubuntu-18.04,   r: 'oldrel-4'}
-          - {os: ubuntu-18.04,   r: 'release', cran: "https://demo.rstudiopm.com/all/__linux__/bionic/latest", locale: 'en_US'}
```
jennybc, line 38: We need to keep this.

.github/workflows/R-CMD-check.yaml
```diff
@@ -44,16 +43,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set locale
-        if: matrix.config.locale == 'en_US'
-        run: |
-          sudo locale-gen en_US
-          echo "LC_ALL=en_US" >> $GITHUB_ENV
-
```
jennybc, line 52: We need to keep this.

.github/workflows/R-CMD-check.yaml
```diff
@@ -44,16 +43,6 @@ jobs:
...
-      - name: Install XQuartz on macOS
-        if: runner.os == 'macOS'
-        run: brew install --cask xquartz
-
```
jennybc, line 56: Off the top of my head, I don't see why this is necessary here in vroom. Can you look at git blame / git history to see if you can discover any reason we (still) need this?

.github/workflows/R-CMD-check.yaml
```diff
@@ -62,19 +51,11 @@ jobs:
...
-      - name: Install system dependencies-2
-        if: runner.os == 'Linux'
-        run: |
-          sudo apt-get install -y libarchive-dev
-
-      - name: Install system dependencies-3
-        if: runner.os == 'macOS'
-        run: |
-          brew install libarchive
-
```
jennybc, line 74: Can you also do some research here to see why / when this was introduced? My hunch is that it was in a period when archive needed to be built from source in GHA. But I don't think this is a constant ongoing need.

_pkgdown.yml
```diff
@@ -6,8 +6,6 @@ template:
   includes:
     in_header: |
       <script defer data-domain="vroom.r-lib.org,all.tidyverse.org" src="https://plausible.io/js/plausible.js"></script>
-  params:
-    ganalytics: UA-115082821-1
```
jennybc, line 9: This doesn't really fit in this PR, but it's not a huge deal.

.github/workflows/R-CMD-check.yaml
sbearrows: Here is where it was originally added

https://github.com/r-lib/vroom/commit/08e483877cfa0a87c45fcb739892c319d349b6aa

But more recently, some similar code from `.github/workflows/pkgdown.yaml` was removed here:

https://github.com/r-lib/vroom/commit/783c3fa5d082e823cf7d586d82f0c56fa0fd125e

Which I think supports that we can remove this?

.github/workflows/R-CMD-check.yaml
sbearrows: That seems right. Here is where it was first added

https://github.com/r-lib/vroom/commit/81afde37c968904c8e5c0caa633fedc7a00920f6

.github/workflows/R-CMD-check.yaml
jennybc: I think it's safe to remove. We also saw CI pass (except for the job that is failing due to the rmarkdown problem) with this deleted, in this PR.

.github/workflows/R-CMD-check.yaml
jennybc: Yeah, I believe we no longer need this. The deletion is good to go.

.github/workflows/R-CMD-check.yaml
```diff
@@ -57,5 +64,3 @@ jobs:
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
-        with:
-          upload-snapshots: true
```
jennybc: We want these lines. I.e. this a new addition from upstream that we want to incorporate.

jennybc (approved):
LGTM

PR #418, opened 2022-03-14
Update NA handling for factors with no explicit levels

### LINKED ISSUE #396: vroom na option being ignored for factors
Hello,
I have an issue using the `na` option in vroom. (I was initially using readr but saw that vroom is being used by readr).

```
test.df = data.frame(a=c("missed","a","b"),b=c(1,2,3),c=c(4,5,6))
write.csv(test.df,"test.csv",row.names = FALSE)
x = vroom::vroom("test.csv",na="missed",col_types=cols(a=col_factor(),b=col_integer(),c=col_integer()))
summary(x)
```

I get the following output:
```
      a           b             c      
 missed:1   Min.   :1.0   Min.   :4.0  
 a     :1   1st Qu.:1.5   1st Qu.:4.5  
 b     :1   Median :2.0   Median :5.0  
            Mean   :2.0   Mean   :5.0  
            3rd Qu.:2.5   3rd Qu.:5.5  
            Max.   :3.0   Max.   :6.0
```

My understanding is that 'missed' should have been read as a missing value. I apologize if I'm misunderstanding the option.

sbearrows:
Closes #396
Closes https://github.com/tidyverse/readr/issues/1292

### DIFF
--- a/src/vroom_fct.cc
+++ b/src/vroom_fct.cc
@@ -73,9 +73,9 @@ cpp11::integers read_fct_implicit(vroom_vec_info* info, bool include_na) {
           if (na_level == NA_INTEGER) {
             na_level = max_level++;
             levels.push_back(NA_STRING);
-            out[i++] = na_level;
             level_map[str.str()] = na_level;
           }
+          out[i++] = na_level;
         } else {
           out[i++] = NA_INTEGER;
         }
--- a/src/vroom_fct.cc
+++ b/src/vroom_fct.cc
@@ -68,15 +68,17 @@ cpp11::integers read_fct_implicit(vroom_vec_info* info, bool include_na) {
     if (val != level_map.end()) {
       out[i++] = val->second;
     } else {
-      if (include_na && matches(str, nas)) {
-        if (na_level == NA_INTEGER) {
-          na_level = max_level++;
-          levels.push_back(NA_STRING);
-          out[i++] = na_level;
-          level_map[str.str()] = na_level;
+      if (matches(str, nas)) {
+        if (include_na) {
+          if (na_level == NA_INTEGER) {
+            na_level = max_level++;
+            levels.push_back(NA_STRING);
+            out[i++] = na_level;
+            level_map[str.str()] = na_level;
+          }
+        } else {
+          out[i++] = NA_INTEGER;
         }
-      } else if (!include_na && matches(str, nas)) {
-        out[i++] = NA_INTEGER;
       } else {
         out[i++] = max_level;
         level_map[str.str()] = max_level++;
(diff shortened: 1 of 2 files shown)


src/vroom_fct.cc
```diff
@@ -75,6 +75,8 @@ cpp11::integers read_fct_implicit(vroom_vec_info* info, bool include_na) {
           out[i++] = na_level;
           level_map[str.str()] = na_level;
         }
+      } else if (!include_na && matches(str, nas)) {
+        out[i++] = NA_INTEGER;
```
DavisVaughan, line 79: I wonder if the `include_na` and `!include_na` branches could be combined. With a structure like:

```cpp
if (matches(str, nas)) {
  if (include_na) {

  } else {

  }
} else {

}
```

jennybc: I like @DavisVaughan 's suggestion. Rewriting it that way makes it easier to see that maybe you've got an unhandled case, which is leading to the malformed factor. As it stands, what happens when `include_na` is `TRUE` and you encounter an `NA` string after the first time, i.e. after you've created the `NA` level?

I think it will end up more like this:

```
    if (matches(str, nas)) {
        if (include_na) {
          if (na_level == NA_INTEGER) {
            na_level = max_level++;
            levels.push_back(NA_STRING);
            level_map[str.str()] = na_level;
          }
          out[i++] = na_level;
        } else {
          out[i++] = NA_INTEGER;
        }
      } else {
        out[i++] = max_level;
        level_map[str.str()] = max_level++;
        levels.push_back(
          info->locale->encoder_.makeSEXP(str.begin(), str.end(), false));
      }
    }
```

sbearrows: I agree that Davis's edits make it look cleaner so I'm making those now and will push soon.

The loop checks a few lines up if a value already exists

```
    auto val = level_map.find(str.str());
    if (val != level_map.end()) {
      out[i++] = val->second;
    }
```
so `out` gets set further up if a level already exists (including for `NA`s).

jennybc: But there can be multiple NA strings and that is the case where the malformed factor is being created in the tests.

sbearrows: Ah yes I see! Sorry, I misunderstood the context.

src/vroom_fct.cc
```diff
@@ -73,9 +73,9 @@ cpp11::integers read_fct_implicit(vroom_vec_info* info, bool include_na) {
           if (na_level == NA_INTEGER) {
             na_level = max_level++;
             levels.push_back(NA_STRING);
-            out[i++] = na_level;
             level_map[str.str()] = na_level;
```
jennybc, line 76: I feel like this may need to come out of the `if(na_level == NA_INTEGER)` also? It feels like we aren't adding additional NA strings to the `level_map`, after the first one.

jennybc: IIUC that won't affect correctness, but I guess it could make things slower. In any case, the intention of `level_map` certainly seems to be documenting everything we've seen so far on our "level discovery" journey.

sbearrows: No you're right, it should definitely come out too. Thank you!

jennybc (approved):
LGTM!

PR #428, opened 2022-04-13
`vroom_format()` handles empty data frames

### LINKED ISSUE #425: `format_csv` segfaults with empty data frame
Hi readr team,

I noticed that passing a data.frame with 0 rows/cols causes a segfault. I think it would be nice to check for this, perhaps in `format_delim` right after the call to check that `is.data.frame(x)` and return an empty string in that case, since in some cases (like mine) I'm using it programmatically to deal with incoming data and there might be genuine empty records, so downstream processing can take care of the empty csv.

I am using readr version 2.1.2 with R version 4.0.4.

```r
> library(readr)
> readr::format_csv(data.frame())

 *** caught segfault ***
address 0xffffffffffffffff, cause 'memory not mapped'
```

sbearrows:
Closes #425 

`vroom_format()` and `readr::format_csv()` return the empty string now.

### DIFF
--- a/R/vroom_write.R
+++ b/R/vroom_write.R
@@ -126,7 +126,7 @@ vroom_format <- function(x, delim = "\t", eol = "\n", na = "NA", col_names = TRU
 
   stopifnot(is.data.frame(x))
 
-  if (length(x) == 0) {
+  if (NCOL(x) == 0) {
     return("")
   }
 
--- a/R/vroom_write.R
+++ b/R/vroom_write.R
@@ -124,6 +124,12 @@ vroom_format <- function(x, delim = "\t", eol = "\n", na = "NA", col_names = TRU
                          bom = FALSE,
                          num_threads = vroom_threads()) {
 
+  stopifnot(is.data.frame(x))
+
+  if (length(x) == 0) {
+    return("")
+  }
+
   quote <- match.arg(quote)
   escape <- match.arg(escape)
 
--- a/src/vroom_write.cc
+++ b/src/vroom_write.cc
@@ -322,11 +322,13 @@ std::vector<char> get_header(
       out.push_back(delim);
     }
   }
-  if (delim != '\0') {
-    out.pop_back();
-  }
-  for (auto c : eol) {
-    out.push_back(c);
+  if(out.size() != 0) {
+    if (delim != '\0') {
+      out.pop_back();
+    }
+    for (auto c : eol) {
+      out.push_back(c);
+    }
   }
   return out;
 }
--- a/tests/testthat/test-vroom_write.R
+++ b/tests/testthat/test-vroom_write.R
@@ -209,13 +209,15 @@ test_that("vroom_format handles empty data frames", {
   df <- data.frame(a = 1:2, b = 2:3)
   df <- df[0, ]
   expect_equal(vroom_format(df), "a\tb\n")
+})
 
+test_that("vroom_write() / vroom_read() roundtrips an empty data frame", {
   df <- tibble::tibble()
   t <- tempfile(fileext = ".csv")
   on.exit(unlink(t))
 
-  vroom::vroom_write(df, t)
-  expect_equal(vroom::vroom(t, col_types = list()), df)
+  vroom_write(df, t)
+  expect_equal(vroom(t, show_col_types = FALSE), df)
 })
 
 test_that("vroom_write(append = TRUE) works with R connections", {
--- a/tests/testthat/test-vroom_write.R
+++ b/tests/testthat/test-vroom_write.R
@@ -202,6 +202,11 @@ test_that("vroom_write equals the same thing as vroom_format", {
   expect_equal(readChar(tf, file.info(tf)$size), vroom_format(df))
 })
 
+test_that("vroom_format handles empty data frames", {
+  df <- data.frame()
+  expect_equal(vroom_format(df), "")
+})
+
 test_that("vroom_write(append = TRUE) works with R connections", {
   df <- data.frame(x = 1, y = 2)
   f <- tempfile(fileext = ".tsv.gz")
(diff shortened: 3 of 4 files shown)


R/vroom_write.R
```diff
@@ -124,6 +124,12 @@ vroom_format <- function(x, delim = "\t", eol = "\n", na = "NA", col_names = TRU
                          bom = FALSE,
                          num_threads = vroom_threads()) {
 
+  stopifnot(is.data.frame(x))
+
+  if (length(x) == 0) {
```
jennybc, line 129: I think we need to consider the case where there are columns, but no rows. I guess we return the header string in that case? This scenario may need some analysis, i.e. if readr 1e does something in particular, what does base R do, etc.

``` r
(x <- data.frame(a = 1:2, b = 3:4))
#>   a b
#> 1 1 3
#> 2 2 4
x[0, ]
#> [1] a b
#> <0 rows> (or 0-length row.names)
length(x[0, ])
#> [1] 2
```

I think we also need to think carefully through the return value. For example, `character()` seems worth considering for the 0-columns case.

jennybc:
Even though you are fixing this on the R side, I still expect to see some sort of robustification in the compiled code to prevent the segfault. There has to be some sort of missing check.

R/vroom_write.R
sbearrows: In readr edition 1, `format_csv()` returns an empty string given an empty data frame. If the data frame only has column names, it returns those names. I'm not sure about base R yet. 


``` r
library(readr)

# empty data frame returns empty string

with_edition(1, format_csv(data.table::data.table()))
#> [1] ""

x <- data.frame(a = "", b = "")

x <- x[0, ]

# data frame with columns only returns column names

with_edition(1, format_csv(x))
#> [1] "a,b\n"
```

<sup>Created on 2022-04-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

sbearrows:
Looks like with readr edition 1, there are no problems with writing and reading empty data frames.

``` r
t <- tempfile(fileext = ".csv")
df <- tibble::tibble()
readr::with_edition(1, readr::write_csv(df, t))
readr::with_edition(1, readr::read_csv(t))
#> # A tibble: 0 × 0
```

<sup>Created on 2022-04-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

But base R aborts after a few lines of an empty file.

``` r
t <- tempfile(fileext = ".csv")
df <- tibble::tibble()
write.csv(df, t)
df_empty <- read.csv(t)
#> Error in read.table(file = file, header = header, sep = sep, quote = quote, : first five rows are empty: giving up
```

<sup>Created on 2022-04-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

But it looks like this error is only happening with `vroom_format` and not with `vroom_write()` because there is a check for `vroom_write()` here https://github.com/tidyverse/vroom/blob/4577a99c480abc5f249f923bd043195db32f3319/R/vroom_write.R#L59-L65

R/vroom_write.R
jennybc: OK I realize now that the 0-rows, >0-columns case was already covered, I just needed to ponder it.

As we just discussed, I think I'd change this from `length()` to `NCOL()` just on principle, to match what's done just above in `vroom_write()`.

src/vroom_write.cc
```diff
@@ -322,11 +322,13 @@ std::vector<char> get_header(
...
-  }
-  for (auto c : eol) {
-    out.push_back(c);
+  if(out.size() != 0) {
+    if (delim != '\0') {
+      out.pop_back();
+    }
+    for (auto c : eol) {
+      out.push_back(c);
+    }
```
jennybc, line 331: Just recording that we've both seen that this is actually the only fix necessary, but we think it's a good idea to do something on the R side as well.

tests/testthat/test-vroom_write.R
```diff
@@ -202,6 +202,22 @@ test_that("vroom_write equals the same thing as vroom_format", {
...
+  expect_equal(vroom_format(df), "a\tb\n")
+
+  df <- tibble::tibble()
+  t <- tempfile(fileext = ".csv")
+  on.exit(unlink(t))
+
+  vroom::vroom_write(df, t)
+  expect_equal(vroom::vroom(t, col_types = list()), df)
+})
+
```
jennybc, line 220: Tests look good.

I propose splitting into two tests, because one is about `vroom_format()` and the other is about `vroom_write()`.

Like this:

``` r
test_that("vroom_format handles empty data frames", {
  df <- data.frame()
  expect_equal(vroom_format(df), "")

  df <- data.frame(a = 1:2, b = 2:3)
  df <- df[0, ]
  expect_equal(vroom_format(df), "a\tb\n")
})

test_that("vroom_write() / vroom_read() roundtrips an empty data frame", {
  df <- tibble::tibble()
  t <- tempfile(fileext = ".csv")
  on.exit(unlink(t))

  vroom_write(df, t)
  expect_equal(vroom(t, show_col_types = FALSE), df)
})
```

Two other changes I made:

* I'm not qualifying anything with `vroom::`. I'm not sure why we see that elsewhere in this file, but I'm assuming it's inessential weirdness.
* I'm using `show_col_types = FALSE` instead of `col_types = list()` (which I realize is used elsewhere), just because it feels like it does *less* and that seems like a good idea. If the goal is just to suppress col spec printing, let's do as little as possible to suppress the printing.

Basically, these are both things I don't see a need to change elsewhere, but we might as well write new tests in the way that makes the most sense to us.

jennybc:
Also remember to add a NEWS bullet.

tests/testthat/test-vroom_write.R
sbearrows: I like your suggestions and I updated the tests and add a NEWS bullet. If everything else looks good I think this is ready.

R/vroom_write.R
sbearrows: Oh gotta do this still!

jennybc (approved):

PR #432, opened 2022-04-27
Column iterators check for non-empty rows

### LINKED ISSUE #430: Segfault when reading multiples and first file has no data rows
Opening separate issue for something observed in https://github.com/tidyverse/vroom/issues/425#issuecomment-1092014807

sbearrows:
Closes #430

Column iterators should check the number of rows when reading in multiple data sources. This prevents bad memory access when the first data source is header-only with 0 rows, but subsequent data has at least one row. Now, column iterators start at the first instances of a non-empty data source.


tests/testthat/test-multi-file.R
```diff
@@ -111,6 +111,29 @@ test_that("vroom works if a file contains no data", {
   expect_equal(res, tibble::tibble(A = 1, B = 2))
 })
 
+test_that("vroom works if some files contain no data, regardless of order (#430)", {
```
jennybc, line 114: I would love to see these files have self-explanatory names, like "header-but-no-data.csv" or "header-and-one-data-row.csv".

I realize this might force you to tweak some other tests that use these same files, but it makes the intent of each test so much more clear.

jennybc (commented):
This looks very promising!

I made a few suggestions that enhance future understandability.

I know you're still doing some informal checks with alternative data sources, i.e. non-files. I guess we'll decide whether to add more tests based on what you find.

tests/testthat/test-multi-file.R
jennybc: Alternatively, if you don't want to touch any other tests, create the exact files you need for this, with good names. These are tiny files, so there's no need to worry about adding files of this size. If/when we touch other tests that could use these new, self-documenting test files, we can switch them over.

And actually: I would probably just inline the creation (and destruction) of these test files. The combination of explicit construction and self-explaining file names would make the intent of each expectation crystal clear.

In this case, I would use `withr::local_tempfile()` to create the test files.

jennybc:
LGTM!

For posterity's sake, I want to record an observation we've had based on fixing this bug and one other recent one: When a file ends earlier-than-expected, it results in an index that has to be used with care. I.e. naive use of the index can lead you to dereference something that basically points past the end of the file, resulting in segfaults.

Here and in #429, we're adding an *ad hoc* guard for concrete edge cases. But if we see more bugs of this nature, we have to consider if something should change in the index-building process (?). Should we be creating indexes that are basically degenerate at certain positions? However, at this point, our understanding of the whole design isn't robust enough to do something like that.

jennybc (approved):
I'm approving this. But before you merge, I think you should add a bullet to the NEWS.

PR #434, opened 2022-05-05
Better handling of non-ascii filepaths

### LINKED ISSUE #394: File path encoding problem on Windows

jennybc:
Closes #394, closes #402, closes #403 (previous explorations)

There's a bit more going on here than we originally thought and the solution isn't quite as simple as "UTF-8 everywhere" (but it's close).

First, the file from the original issue (https://github.com/tidyverse/readr/issues/1345) has a path containing non-ascii characters, but what's easier to miss is that the file also lacks a trailing newline. The failure cascade looks like this:

[`has_trailing_newline()`](https://github.com/tidyverse/vroom/blob/829ba962b8f263e0833d8905376a7cb05bea9a70/src/vroom.cc#L95-L110) implicitly expects that the path is in the native encoding and, if it's not, can potentially fail to find an existing file. When  that happens, it unconditionally reports `TRUE` (yes, there is a trailing new line), which is wrong in this case. A file that lacks a trailing newline should be routed through the connection logic, but, in this case is read as a "regular file", which fails and results in an empty tibble.

So yes the problem is the file path encoding, but it first raises its head via `has_trailing_newline()`, not the main file-reading step.

Second, by exploring various solutions, I discovered that blindly applying `enc2utf8()` is too simple, since vroom tests itself on linux in a non-UTF-8 locale. Hence we do `enc2utf8()` on Windows and `enc2native()` otherwise. This effort did reveal a pre-existing file path problem on this build that I ultimately decided to not fix, in the "file ends with a newline" case. I think this OS-dependent approach may be currently irrelevant in vroom, because of the mio problem I show below, but might be worth it in other settings.

For normal files, on linux, in a non-UTF-8 locale, with a multi-byte file path, we get an error from mio itself, i.e. the memory mapping fails:

https://github.com/tidyverse/vroom/runs/6315284023?check_suite_focus=true#step:7:182

```
> test_check("vroom")
mapping error: No such file or directory
[ FAIL 2 | WARN 0 | SKIP 2 | PASS 1065 ]
```

I am content to `skip()` here and re-consider if an actual user ever needs to do this. There is also evidence that vroom's writing functions are not entirely prepared for this scenario, but the same posture applies.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -25,6 +25,8 @@ jobs:
           - {os: macOS-latest,   r: 'release'}
 
           - {os: windows-latest, r: 'release'}
+          # Use 4.1 to trigger usage of RTools40
+          - {os: windows-latest, r: '4.1'}
           # Use 3.6 to trigger usage of RTools35
           - {os: windows-latest, r: '3.6'}
 
@@ -35,7 +37,8 @@ jobs:
           - {os: ubuntu-18.04,   r: 'oldrel-2'}
           - {os: ubuntu-18.04,   r: 'oldrel-3'}
           - {os: ubuntu-18.04,   r: 'oldrel-4'}
-          - {os: ubuntu-18.04,   r: 'release', cran: "https://demo.rstudiopm.com/all/__linux__/bionic/latest", locale: 'en_US'}
+          # Check with a non-UTF-8 encoding
+          - {os: ubuntu-18.04,   r: 'release', cran: "https://demo.rstudiopm.com/all/__linux__/bionic/latest", locale: 'en_US.ISO-8859-1'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -45,7 +48,8 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set locale
-        if: matrix.config.locale == 'en_US'
+        if: matrix.config.locale == 'en_US.ISO-8859-1'
+        # en_US is an alias for what we want
         run: |
           sudo locale-gen en_US
           echo "LC_ALL=en_US" >> $GITHUB_ENV
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* `vroom()` reads more reliably from filepaths containing non-ascii characters, on Windows and/or in a non-UTF-8 encoding (#394).
+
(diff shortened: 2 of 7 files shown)


DavisVaughan (approved):
Such a hard fought battle

PR #441, opened 2022-05-17
Improve `problems()` output

sbearrows:
Closes https://github.com/tidyverse/readr/issues/1322

The output message from problems does not make it obvious that it requires a data frame so we should make this more explicit. Current behavior looks like this:

``` r
library(vroom)

has_problems <- glue::glue('
x,y
1,2
3.x,4')

vroom(I(has_problems),
      col_types = "dd")
#> Warning: One or more parsing issues, see `problems()` for details
#> # A tibble: 2 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2    NA     4

problems()
#> Error in vroom_materialize(x, replace = FALSE): argument "x" is missing, with no default
```

<sup>Created on 2022-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

It also isn't obvious that `problems()` reports the file line number from the original file and not the data frame row number. It looks like Jim considered this issue already and ruled that it should report the line number in the original file: https://github.com/tidyverse/vroom/issues/298.

To make this more obvious, some improvements were made to the current documentation, but we should also consider tweaking the column names in the output from `problems()`. For example changing `row` to `line` or something similar. This would require a bit more work than what is currently in the PR but would make the source of `problems()` more obvious.

``` r
library(vroom)
has_problems <- glue::glue("
x,y
1,2
3.x,4")

# proposed renaming
problems(
  vroom(I(has_problems),
        col_types = "dd"
  )
) |> dplyr::rename(line = row)
#> Warning: One or more parsing issues, see `problems()` for details
#> # A tibble: 1 × 5
#>    line   col expected actual file                                              
#>   <int> <int> <chr>    <chr>  <chr>                                             
#> 1     3     1 a double 3.x    /private/var/folders/4g/9jcx0hbd6r92152m0qrq64yw0…
```

<sup>Created on 2022-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>


R/problems.R
```diff
@@ -10,7 +10,8 @@
 #'   `FALSE` (the default) the lazy data is first read completely and all
 #'   problems are returned.
 #' @return A data frame with one row for each problem and four columns:
-#'   - row,col - Row and column of problem
+#'   - row,col - Row and column number that caused the problem, referencing the
+#'   original file
```
jennybc, line 14: I suspect we should say something more generic, like "input", to account to connections and inlined/literal inputs.

src/vroom_errors.h
```diff
@@ -89,7 +89,7 @@ class vroom_errors {
       cpp11::sexp warn_call = Rf_lang3(
           warn,
           Rf_mkString(
-              "One or more parsing issues, see `problems()` for details"),
+              "One or more parsing issues, use `problems()` on your data frame for details"),
```
jennybc, line 92: What do you think of "call" (instead of "use")?

I wonder if we should give a concrete demo?

```
One or more parsing issues, call `problems()` on your data frame for details, e.g.:
  dat <- vroom(...)
  problems(dat)
```

It always feels awkward to emit rich messages from C++, so let's try to decide on the content without worrying too much about the mechanics at first.

jennybc (commented):
`problems()` is a great candidate for snapshot testing, ~but I note that vroom doesn't use testthat 3e yet. We'll have to think about whether it's time to opt-in to that.~ *I was wrong, it does.*

src/vroom_errors.h
sbearrows: I think `call` is definitely better and an example would be nice and explicit

R/problems.R
sbearrows: Agree

jimhester:
vroom actually does use testthat 3e (https://github.com/tidyverse/vroom/blob/main/DESCRIPTION#L75) unless I am missing something.

tests/testthat/test-problems.R
```diff
@@ -19,42 +15,34 @@ test_that("problems works for multiple files", {
...
+    problems(v3)
+  })
+
+ expect_snapshot({
+    v4 <- vroom(I("x,y\n1,2,3,4\n"), col_names = FALSE, col_types = "dd", altrep = FALSE)
+    problems(v4)
+  })
 })
 
 test_that("parsing problems are shown for all datatypes", {
```
sbearrows, line 48: I'm still thinking about how to use snapshots here. My first take was replacing the `expect_warning()` and `expect_equal()` with `expect_snapshot()`but this felt repetitive because of the looping.

sbearrows:
@jennybc this is ready for another review

src/vroom_errors.h
```diff
@@ -83,15 +85,12 @@ class vroom_errors {
...
-          Rf_mkString(
-              "One or more parsing issues, see `problems()` for details"),
-          Rf_mkString("vroom_parse_issue"));
-      Rf_eval(warn_call, R_EmptyEnv);
+      static auto cli_warn = cpp11::package("cli")["cli_warn"];
+      cpp11::strings bullets({
+        "w"_nm = "One or more parsing issues, call `problems()` on your data frame for details, e.g.:",
+        " "_nm = "dat <- vroom(...)",
+        " "_nm = "problems(dat)"});
+      cpp11::sexp warn_call = cli_warn(bullets, "class"_nm = "vroom_parse_issue", ".envir"_nm = R_EmptyEnv);
```
jennybc, line 93: Since `warn_for_errors()` returns `void`, it seems best to simplify this.

```suggestion
      cli_warn(bullets, "class"_nm = "vroom_parse_issue", ".envir"_nm = R_EmptyEnv);
```

src/vroom_errors.h
```diff
@@ -83,15 +85,12 @@ class vroom_errors {
...
-          Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("rlang")));
-      cpp11::sexp warn_call = Rf_lang3(
-          warn,
-          Rf_mkString(
-              "One or more parsing issues, see `problems()` for details"),
-          Rf_mkString("vroom_parse_issue"));
-      Rf_eval(warn_call, R_EmptyEnv);
+      static auto cli_warn = cpp11::package("cli")["cli_warn"];
+      cpp11::strings bullets({
+        "w"_nm = "One or more parsing issues, call `problems()` on your data frame for details, e.g.:",
```
jennybc, line 90: Since we have the full power of cli at our disposal, we can lean in to it even more. This way we're requesting to style `problems()` "like a function" and leaving the details to cli:

```suggestion
        "w"_nm = "One or more parsing issues, call {.fun problems} on your data frame for details, e.g.:",
```

jennybc (commented):
Overall, looks good!

I think there are too many snapshot tests now. I.e. we've crossed the line into storing a lot of snapshots that aren't actually doing much for us -- they're mostly redundant with one another. Since the warning is static, one dedicated snapshot test for this warning message is enough.

Then the remaining expectations can go back to `expect_warning()`, probably with the addition of `class = "vroom_parse_issue"`, to make sure we're catching the warning we expect. That feels like the right compromise.

tests/testthat/test-problems.R
```diff
@@ -11,16 +11,15 @@ test_that("problems with data parsing works for single files", {
...
   out2 <- file.path(tempdir(), "out2.txt")
   on.exit(unlink(c(out1, out2)))
 
   writeLines("x,y\n1,2\n1,1.x\n2,2", out1)
   writeLines("x,y\n3.x,4\n1,2\n2,2", out2)
 
-  expect_warning(
-    x <- vroom(c(out1, out2), delim = ",", col_types = "dd", altrep=F)
+  expect_snapshot(
+    x <- vroom(c(out1, out2), delim = ",", col_types = "dd", altrep = F)
```
jennybc, line 22:
```suggestion
    x <- vroom(c(out1, out2), delim = ",", col_types = "dd", altrep = FALSE)
```

Seeing this makes me this we should probably subject the entire file to styler, since we're touching it.

tests/testthat/test-problems.R
```diff
@@ -78,18 +77,18 @@ test_that("parsing problems are shown for all datatypes", {
     res <- vroom(I("x\nxyz\n"), delim = ",", col_types = list(type), altrep = TRUE)
 
     # This calls the type_Elt function
-    expect_warning(res[[1]][[1]], "One or more parsing issues")
```
jennybc, line 81: This is an example of where stating the expected class is going to be nice.

src/vroom_errors.h
```diff
@@ -83,15 +85,12 @@ class vroom_errors {
...
-          Rf_mkString(
-              "One or more parsing issues, see `problems()` for details"),
-          Rf_mkString("vroom_parse_issue"));
-      Rf_eval(warn_call, R_EmptyEnv);
+      static auto cli_warn = cpp11::package("cli")["cli_warn"];
+      cpp11::strings bullets({
+        "w"_nm = "One or more parsing issues, call {.fun problems} on your data frame for details, e.g.:",
+        " "_nm = "dat <- vroom(...)",
+        " "_nm = "problems(dat)"});
+      cli_warn(bullets, "class"_nm = "vroom_parse_issue", ".envir"_nm = R_EmptyEnv);
```
jennybc, line 93: Upon reading more of the history of this function, it looks like there's a specific reason to _not_ use `cpp11::package`. But I still think we can switch to `cli::cli_warn()` (as opposed to `rlang::warn()`).

https://github.com/tidyverse/vroom/commit/984a3e5e37e124feacfec3d184dbeb02eb1145c4

DavisVaughan: I can somewhat explain this. If you use `cpp11::package()` to call `warn()`, which involves the unwind protect, then if a user tries to catch the classed warning with `withCallingHandlers()` and tries to promote the warning to an error, then it will actually crash R.

I've set up a draft PR to demonstrate this: https://github.com/DavisVaughan/vroom/pull/1

Here is an example where the user might catch the classed warning and promote it to an error. This works with CRAN vroom correctly, as you see here:

``` r
library(vroom)

make_warning <- function() {
  x <- vroom(
    I("a,b\n1.0,x\n"), 
    delim = ",", 
    col_types = "dd"
  )
  
  # Warning happens at print time so force a print
  print(x)
}

# This fails hard if we unwind protect the warning (aborts RStudio)
# - Try to throw error after catching the warning
withCallingHandlers(
  expr = make_warning(),
  vroom_parse_issue = function(cnd) {
    rlang::abort("oh no")
  }
)
#> Error:
#> ! oh no
```

If you run this with the PR installed, then it will crash R. I don't quite understand why, but it seems to definitely be related to the unwind protect.

It seems to be only be related to error calls. I tried a few other variations where I called `print()` or `warn()` in the condition handler, and I also tried muffling the warning, and that all seemed to work fine even with the unwind protect. I have included those examples below (This was run with the PR installed).

<details>

``` r
library(vroom)

make_warning <- function() {
  x <- vroom(
    I("a,b\n1.0,x\n"), 
    delim = ",", 
    col_types = "dd"
  )
  
  # Warning happens at print time so force a print
  print(x)
}

# Works fine:
# - Print some stuff
# - Muffle the condition
withCallingHandlers(
  expr = make_warning(),
  vroom_parse_issue = function(cnd) {
    print("oh no")
    rlang::cnd_muffle(cnd)
  }
)
#> [1] "oh no"
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1    NA

# Works fine:
# - Print some stuff
# - Don't muffle the condition
withCallingHandlers(
  expr = make_warning(),
  vroom_parse_issue = function(cnd) {
    print("oh no")
  }
)
#> [1] "oh no"
#> Warning: One or more parsing issues, see `problems()` for details
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1    NA

# Works fine:
# - Throw a new warning
# - Muffle the condition
withCallingHandlers(
  expr = make_warning(),
  vroom_parse_issue = function(cnd) {
    rlang::warn("oh no")
    rlang::cnd_muffle(cnd)
  }
)
#> Warning: oh no
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1    NA

# Works fine:
# - Throw a new warning
# - Don't muffle the condition
withCallingHandlers(
  expr = make_warning(),
  vroom_parse_issue = function(cnd) {
    rlang::warn("oh no")
  }
)
#> Warning: oh no
#> One or more parsing issues, see `problems()` for details
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1    NA
```

<sup>Created on 2022-05-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

</details>

If it were me, this would prompt me to include a test for `test_that("can promote the vroom parse warning to an error")` that uses `withCallingHandlers()` like in my example above. That would ensure we don't accidentally screw that feature up in the future.

DavisVaughan: Regarding the `static` in front of:

```cpp
static auto warn = Rf_findFun(
           Rf_install("warn"),
           Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("rlang")));
```

That means that the initialization of `warn` is only run once. It may be somewhat expensive to look up `warn()` from the rlang namespace (probably not for this case), so this ensures that `warn` is only looked up once, and is stored until R is shut down.

https://stackoverflow.com/questions/5567529/what-makes-a-static-variable-initialize-only-once

jennybc: Thanks @DavisVaughan for the additional analysis.

I agree adding that test is a good idea.

jennybc (approved):

PR #442, opened 2022-05-20
Remove the width from the taylor gif

jimhester:
Having only 30% is making it look way to small on GitHub, and it seems
to still look right on the pkgdown site as well without an explicit
width.


jennybc:
Welcome back, Jim! I was hoping to see a giant PR on the C++ side, but we'll take what we can get 🤣

jimhester:
Haha, that would take a bit more time then this ;)

PR #451, opened 2022-07-16
Don’t overwrite existing file if appending an empty data frame

sbearrows:
Fixes tidyverse/readr#1408

Previously, if `vroom_write()` was supplied an empty data frame, vroom created an empty file without checking the value of `append`. But this meant data from the file could be deleted such as in this case.

``` 
library(vroom)
tf <- withr::local_tempfile()
data <- tibble::tibble(a = "1", b = "2", c = "3")

vroom_write(data, file = tf)
vroom(tf, show_col_types = FALSE)
#> # A tibble: 1 × 3
#>       a     b     c
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3

vroom_write(data.frame(), file = tf, append = TRUE)
vroom(tf, show_col_types = FALSE)
#> # A tibble: 0 × 0
```

<sup>Created on 2022-07-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

### DIFF
--- a/R/vroom_write.R
+++ b/R/vroom_write.R
@@ -58,10 +58,9 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names 
   # Standardise path returns a list, but we will only ever have 1 output file.
   file <- standardise_one_path(file, write = TRUE)
 
-  # If there are no columns in the data frame and we aren't appending
-  # just create an empty file and return
-  if (NCOL(x) == 0 && !append) {
-    if (!inherits(file, "connection")) {
+  # If there are no columns in the data frame then return early
+  if (NCOL(x) == 0) {
+    if (!append && !inherits(file, "connection")) {
       file.create(file)
     }
     return(invisible(input))
--- a/R/vroom_write.R
+++ b/R/vroom_write.R
@@ -58,8 +58,9 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names 
   # Standardise path returns a list, but we will only ever have 1 output file.
   file <- standardise_one_path(file, write = TRUE)
 
-  # If there are no columns in the data frame, just create an empty file and return
-  if (NCOL(x) == 0) {
+  # If there are no columns in the data frame and we aren't appending
+  # just create an empty file and return
+  if (NCOL(x) == 0 && !append) {
     if (!inherits(file, "connection")) {
       file.create(file)
     }
--- a/tests/testthat/test-vroom_write.R
+++ b/tests/testthat/test-vroom_write.R
@@ -312,10 +312,10 @@ test_that("vroom_write() does not overwrite file when appending empty data frame
   data <- tibble::tibble(a = "1", b = "2", c = "3")
 
   vroom_write(data, file = tf)
-  first_write <- vroom(tf, show_col_types = FALSE)
+  first_write <- vroom(tf, altrep = FALSE, show_col_types = FALSE)
 
   vroom_write(data.frame(), file = tf, append = TRUE)
-  second_write <- vroom(tf, show_col_types = FALSE)
+  second_write <- vroom(tf, altrep = FALSE, show_col_types = FALSE)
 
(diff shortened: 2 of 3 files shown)


jennybc:
I recommend turning off altrep / lazy reading in this test.

R/vroom_write.R
```diff
@@ -58,8 +58,9 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names
   # Standardise path returns a list, but we will only ever have 1 output file.
   file <- standardise_one_path(file, write = TRUE)
 
-  # If there are no columns in the data frame, just create an empty file and return
-  if (NCOL(x) == 0) {
+  # If there are no columns in the data frame and we aren't appending
+  # just create an empty file and return
+  if (NCOL(x) == 0 && !append) {
```
jennybc, line 63: What about the case where `NCOL(x) == 0 && append`?
It feels like that should still be an early return.

sbearrows:
@jennybc I believe this is ready for another look!

R/vroom_write.R
```diff
@@ -58,9 +58,9 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names
   # Standardise path returns a list, but we will only ever have 1 output file.
   file <- standardise_one_path(file, write = TRUE)
 
-  # If there are no columns in the data frame, just create an empty file and return
+  # If there are no columns in the data frame then return early
```
jennybc, line 61:
```suggestion
```

I think the fact that we're returning early doesn't need a comment.

R/vroom_write.R
```diff
@@ -58,9 +58,9 @@ vroom_write <- function(x, file, delim = '\t', eol = "\n", na = "NA", col_names
...
   if (NCOL(x) == 0) {
-    if (!inherits(file, "connection")) {
+    if (!append && !inherits(file, "connection")) {
```
jennybc, line 63: I think this has become the comment-worthy logic:

```suggestion
    if (!append && !inherits(file, "connection")) {
      # if file already exists, it is overwritten with an empty file!
```

I am mildly surprised `vroom_write()` doesn't expose an argument or messaging around overwriting an existing file. But since that's how we found it, that's how we'll leave it.

tests/testthat/test-vroom_write.R
```diff
@@ -306,3 +306,23 @@ test_that("na argument modifies how missing values are written", {
...
+    c = "3"
+  )
+
+  vroom_write(data, file = tf, delim = ",")
+  vroom_write(data.frame(), file = tf, append = TRUE, delim = ",")
+
+  expect_equal(
+    vroom_lines(tf,
+      altrep = FALSE,
+      skip_empty_rows = FALSE
```
jennybc, line 324: I can't see why we need to specify `skip_empty_rows`.

If we don't, let's remove it. In which case, the expectation can be come more compact.

sbearrows: We don't need it. I just wanted to make sure it wasn't writing an empty line when `vroom_write(data.frame(), file = file, append = TRUE)`

jennybc: I think if `vroom_write(data.frame(), file = file, append = TRUE)` wrote an empty line, that would be a bug and we would want to know about it. So I think the test is better without it (which I see has happened now).

jennybc: Nitpick:

> let's remove it. In which case, the expectation can be come more compact.

There's still this.

jennybc (approved):
I made the expectation in the new test more similar to other in this file / package.

I added a NEWS bullet.

R/vroom_write.R
sbearrows: Notes: Since we weren't returning early when `NCOL(x) == 0 && append` (ie writing an empty dataframe to a file with data already present), `vroom_write_` was being passed an empty data frame, which caused R to crash on Windows with R 4.1 and 3.6 (both of which are 32-bit builds).

PR #452, opened 2022-07-18
Update test for `unwind::protect` error in `vroom_errors::warn_for_errors()`

### LINKED ISSUE #445: Update test for `problems()` `unwind::protect` error
The test in question: https://github.com/tidyverse/vroom/blob/f21f3143f659745b06452ec8235819ca098fb2b0/tests/testthat/test-problems.R#L149-L172

In a PR to update the warning message from `problems()` we attempted to use `cpp11::package()` to call `cli::cli_warn()`. For a number of reasons this isn't a great idea and can actually crash R https://github.com/tidyverse/vroom/pull/441#discussion_r883611090

At the time we wrote the test, we didn't fully understand the circumstances that caused the crash. Now that we are more aware of how it occurs, we should update the test to reflect this information.

sbearrows:
Fixes #445 

This might be a bit redundant/circular, but I'm going to relink @DavisVaughan issue post about this because I found it really useful for refreshing my memory on the problem

https://github.com/r-lib/cpp11/issues/274

### DIFF
--- a/tests/testthat/test-problems.R
+++ b/tests/testthat/test-problems.R
@@ -158,9 +158,9 @@ test_that("can promote vroom parse warning to error", {
 
     # Trigger vroom parse warning while inside R's internal C code for `[` and ensure it doesn't crash R.
     # `[` -> R's C function `do_subset()` -> ALTREP calls `vroom::real_Elt()` -> `vroom::warn_for_errors()`
-    # `warn_for_errors()` warns with base R's machinery rather than cpp11's to avoid calling
-    # `unwind_protect()` (which throws on longjmp, i.e. on `abort()`) while inside R's internal
-    # C code (which doesn't catch C++ exceptions).
+    # To avoid calling `cpp11::unwind_protect()` (which throws on longjmp, i.e. on `abort()`) while inside
+    # R's internal C code (which doesn't catch C++ exceptions), `vroom::warn_for_errors()` warns
+    # with base R's machinery rather than cpp11's
     # https://github.com/r-lib/cpp11/issues/274
     # https://github.com/tidyverse/vroom/pull/441#discussion_r883611090
     x$a[1]
--- a/tests/testthat/test-problems.R
+++ b/tests/testthat/test-problems.R
@@ -150,13 +150,15 @@ test_that("problems return the proper row number", {
 test_that("can promote vroom parse warning to error", {
   make_warning <- function() {
     x <- vroom(
-      I("a,b\n1.0,x\n"),
+      I("a\nx\n"),
       delim = ",",
-      col_types = "dd"
+      col_types = "d",
+      altrep = TRUE
     )
 
-    # Warning happens at print time so force a print
-    print(x)
+    # trigger the warning using R's internal [ code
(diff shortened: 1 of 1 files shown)


tests/testthat/test-problems.R
```diff
@@ -150,13 +150,15 @@ test_that("problems return the proper row number", {
...
-      col_types = "dd"
+      col_types = "d",
+      altrep = TRUE
     )
 
-    # Warning happens at print time so force a print
-    print(x)
+    # do something to trigger the warning using R's internal C code
+    # https://github.com/r-lib/cpp11/issues/274
+    x$a[1]
```
DavisVaughan, line 161: After rereading my in depth issue, how do you feel about this block of comments?

I think you could remove all the other comments in and around this test-that block, and put this block of comments right after the `test_that()` line and before the `make_warning <- ` line

```suggestion
    # Trigger vroom parse warning while inside R's internal C code for `[` and ensure it doesn't crash R.
    # `[` -> R's C function `do_subset()` -> ALTREP calls `vroom::real_Elt()` -> `vroom::warn_for_errors()`
    # `warn_for_errors()` warns with base R's machinery rather than cpp11's to avoid calling
    # `unwind_protect()` (which throws on longjmp, i.e. on `abort()`) while inside R's internal
    # C code (which doesn't catch C++ exceptions).
    # https://github.com/r-lib/cpp11/issues/274
    # https://github.com/tidyverse/vroom/pull/441#discussion_r883611090
    x$a[1]
```

sbearrows:
@jennybc This is ready for you to review

tests/testthat/test-problems.R
```diff
@@ -150,13 +150,20 @@ test_that("problems return the proper row number", {
...
+      altrep = TRUE
     )
 
-    # Warning happens at print time so force a print
-    print(x)
+    # Trigger vroom parse warning while inside R's internal C code for `[` and ensure it doesn't crash R.
+    # `[` -> R's C function `do_subset()` -> ALTREP calls `vroom::real_Elt()` -> `vroom::warn_for_errors()`
+    # To avoid calling `cpp11::unwind_protect()` (which throws on longjmp, i.e. on `abort()`) while inside
+    # R's internal C code (which doesn't catch C++ exceptions), `vroom::warn_for_errors()` warns
+    # with base R's machinery rather than cpp11's
```
jennybc, line 163: > `vroom::warn_for_errors()` warns with base R's machinery rather than cpp11's

This review prompted me to read :point_up_2: for the first time in a long time and then compare it to the source for `warn_for_error()`:

https://github.com/tidyverse/vroom/blob/f21f3143f659745b06452ec8235819ca098fb2b0/src/vroom_errors.h#L85-L103

With fresh eyes, I feel like we can refine this to be even more precise. I'd say we're warning with cli and we're using base R machinery to access the relevant R function, as opposed to accessing it via cpp11 way.

jennybc (approved):
The update to the test looks good.

I'm pre-approving, but am also suggesting a change to this load-bearing comment, which is a key piece of documentation for something quite subtle.

PR #454, opened 2022-08-04
fix issue with empty date and datetime columns

### LINKED ISSUE #453: empty column of type date or date-time gives error in R version 3.6.3 with skip = 1
In R version 3.6.3, reading data with a date or date-time column and a header when also skipping the first line throws an error.
```
$ Rscript -e "vroom::vroom(I('date\n'), col_names = 'date', col_types = 'D', skip = 1L)"
Error in as.Date.numeric(double()) : 'origin' must be supplied
Calls: <Anonymous> ... collector_value.collector_date -> as.Date -> as.Date.numeric
Execution halted
```
```
$ Rscript -e "vroom::vroom(I('dt\n'), col_names = 'dt', col_types = 'T', skip = 1L)"
Error in as.POSIXct.numeric(double()) : 'origin' must be supplied
Calls: <Anonymous> ... collector_value.collector_datetime -> as.POSIXct -> as.POSIXct.numeric
Execution halted
```

This is due to the implementation of the numeric converters. E.g.:
```
$ Rscript -e base::as.Date.numeric
function (x, origin, ...) 
{
    if (missing(origin)) 
        stop("'origin' must be supplied")
    as.Date(origin, ...) + x
}
```

In newer versions, there is no error, due to the improved implementation:
```
$ Rscript -e base::as.Date.numeric
function (x, origin, ...) 
{
    if (missing(origin)) {
        if (!length(x)) 
            return(.Date(numeric()))
        if (!any(is.finite(x))) 
            return(.Date(x))
        stop("'origin' must be supplied")
    }
    as.Date(origin, ...) + x
}
```

I am writing a small fix with tests.

philaris:
Fixes #453

In older versions of R, like 3.6.3, there is an error when reading an empty date or date-time column with skip = 1.

```
vroom::vroom(I('date\n'), col_names = 'date', col_types = 'D', skip = 1L)
Error in as.Date.numeric(double()) : 'origin' must be supplied
Calls: <Anonymous> ... collector_value.collector_date -> as.Date -> as.Date.numeric
```
```
vroom::vroom(I('dt\n'), col_names = 'dt', col_types = 'T', skip = 1L)
Error in as.POSIXct.numeric(double()) : 'origin' must be supplied
Calls: <Anonymous> ... collector_value.collector_datetime -> as.POSIXct -> as.POSIXct.numeric
```

This commit fixes the error by using a character() vector instead of a double() vector.

### DIFF
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -515,10 +515,10 @@ collector_value.collector_logical <- function(x, ...) { logical() }
 collector_value.collector_factor <- function(x, ...) { factor() }
 
 #' @export
-collector_value.collector_datetime <- function(x, ...) { as.POSIXct(double()) }
+collector_value.collector_datetime <- function(x, ...) { as.POSIXct(character()) }
 
 #' @export
-collector_value.collector_date <- function(x, ...) { as.Date(double()) }
+collector_value.collector_date <- function(x, ...) { as.Date(character()) }
 
 #' @export
 collector_value.collector_time <- function(x, ...) { hms::hms() }
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -514,11 +514,15 @@ collector_value.collector_logical <- function(x, ...) { logical() }
 #' @export
 collector_value.collector_factor <- function(x, ...) { factor() }
 
+# the more obvious as.POSIXct(double()) doesn't work on R < 4.0
+# https://github.com/tidyverse/vroom/issues/453
 #' @export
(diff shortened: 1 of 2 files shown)


jennybc (commented):
Thanks for catching this and the fix. I find using vctrs a bit more expressive and future-facing, so I tweaked the fix. But it produces the same length-0 columns, of the desired type, which is what really matters.

PR #456, opened 2022-08-12
Roll the ubuntu version

jennybc:
This is about aligning to GitHub's posture towards Ubuntu versions.

In GHA, we're starting to see this for vroom:

> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

We deliberately use an "old-ish" Ubuntu version for `R-CMD-check` workflow, but I think it's time to roll forward.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -31,14 +31,14 @@ jobs:
           - {os: windows-latest, r: '3.6'}
 
           # Use older ubuntu to maximise backward compatibility
-          - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-18.04,   r: 'release'}
-          - {os: ubuntu-18.04,   r: 'oldrel-1'}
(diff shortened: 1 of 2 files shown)


DavisVaughan:
I wonder why the warning doesn't show up in, say, this tidyselect PR that ran today https://github.com/r-lib/tidyselect/runs/7806040037?check_suite_focus=true

DavisVaughan:
Oooh GitHub is going to be fairly aggressive about this change. They are going to use brownout periods to temporarily fail builds still using 18.04, and that is going to start fairly soon https://github.com/actions/runner-images/issues/6002#issue-1325639357

jennybc:
vroom actually had a failure recently because we happened to check during a brownout period.

jennybc:
tidyselect **is** getting the warning but it only appears in this view:

https://github.com/r-lib/tidyselect/actions/runs/2846961029

PR #463, opened 2022-08-30
Don't inherit docs from readr

### LINKED ISSUE #461: Don't inherit docs for vroom from readr
Since much of readr is implemented using vroom, it would make more sense to have readr inherit docs from vroom, rather than vroom inheriting docs from readr. We will need to fill in some gaps but we should also update some of the docs in vroom as some of them can be a bit sparse compared to readr:
https://github.com/tidyverse/readr/blob/85cf1e8e664e2a75eba393f91d89b39060140dab/R/source.R#L3-L17
https://github.com/tidyverse/vroom/blob/339073c4fc5cfdcbeb29ff716d41c2aaaf6bad43/R/vroom.R#L8

sbearrows:
Closes #461

Moving over docs from readr to vroom so that the primary docs live here which we think makes more sense.

Note: this would also fix part of https://github.com/tidyverse/readr/issues/1421 (removing mention of `list()` with `col_select`) but we'd still need to add examples of `col_select()`

### DIFF
--- a/R/vroom.R
+++ b/R/vroom.R
@@ -4,8 +4,18 @@ NULL
 
 #' Read a delimited file into a tibble
 #'
-#' @param file Either a path to a file, a connection, or literal data
-#'    (either a single string or a raw vector).
+#' @param file Either a path to a file, a connection, or literal data (either a
+#'   single string or a raw vector).
+#'
+#'   Files ending in `.gz`, `.bz2`, `.xz`, or `.zip` will be automatically
+#'   uncompressed. Files starting with `http://`, `https://`, `ftp://`, or
+#'   `ftps://` will be automatically downloaded. Remote gz files can also be
+#'   automatically downloaded and decompressed.
+#'
+#'   Literal data is most useful for examples and tests. To be recognised as
+#'   literal data, the input must be either wrapped with `I()`, be a string
+#'   containing at least one new line, or be a vector containing at least one
+#'   string with a new line.
 #' @param delim One or more characters used to delimit fields within a
 #'   file. If `NULL` the delimiter is guessed from the set of `c(",", "\t", " ",
 #'   "|", ":", ";")`.
--- a/R/vroom.R
+++ b/R/vroom.R
@@ -4,26 +4,106 @@ NULL
 
 #' Read a delimited file into a tibble
 #'
-#' @inheritParams readr::read_delim
-#' @param file path to a local file.
+#' @param file Either a path to a file, a connection, or literal data
+#'    (either a single string or a raw vector).
 #' @param delim One or more characters used to delimit fields within a
 #'   file. If `NULL` the delimiter is guessed from the set of `c(",", "\t", " ",
 #'   "|", ":", ";")`.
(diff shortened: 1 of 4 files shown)


jennybc:
I need to you do some narration of this task, so I don't have to reverse engineer the stuff below while reviewing. There's a lot to be checked.

Along these lines:

We're tackling the `vroom()` function.
Currently, `n` arguments are documented via inheritance from `readr::some_fcn()` and `m` arguments are documented here in vroom. Of those, `k` are only relevant to `vroom()` and `j` are relevant to both and are documented in both places. Maybe list the arguments in those categories?

I suggest backing up and starting this PR over and probably working with 1 commit per argument. I don't think we're going to have the appetite for doing this wholesale and it's not the best use of our time before release.

From seeing the first commit here, I suggest we prioritize arguments that *only* work in vroom / readr 2e, like `col_select`. It seems clear the primary docs for such an argument should live in vroom. I don't know how many arguments that applies to (which is what the overview of the task would tell us). Let's find and fix these and then re-assess if this task feels worth it relative to everything else that has to happen.

---

If we decide to keep going, then we'd look at arguments that are relevant and documented in both packages, but there's a big difference in the quality of readr's and vroom's docs. Usually, but not always, the vroom docs are less well developed. Then address, probably by writing new parameter docs by combining the best of the readr and vroom text. Example: https://github.com/tidyverse/vroom/commit/ea2358d0c53ae8895e3d825a074c909d30abd3de.Then we'd put this new text in vroom and inherit that from readr.

Watch out for: arguments where behaviour differs in readr 1e and readr 2e / vroom. We can't harmonize willy-nilly if there's a reason the docs differ. This is why I don't think we should do this for all parameters now.

jennybc:
I'm feeling more optimistic about the wholesale update, so let's proceed with that, but with more narration, so it's easier to feel confident about the changes.

R/vroom.R
```diff
@@ -4,26 +4,116 @@ NULL
 
 #' Read a delimited file into a tibble
 #'
-#' @inheritParams readr::read_delim
-#' @param file path to a local file.
+#' @param file Either a path to a file, a connection, or literal data (either a
```
sbearrows, line 7: The only changes/improvements that have been made are to the `file` path parameter here. Everything else was pulled over, as is, from readr so that `vroom()` doesn't inherit from readr anymore. When rendered, the docs are identical. IMO the scope of this PR is only to stop inheriting docs from readr (and potentially only for `vroom()` for now), not necessarily to improve the docs we have. If we find things along the way that need updating, we can make separate PRs for them.

@jennybc What other context would be helpful? Once we finish this I can start https://github.com/tidyverse/readr/issues/1421
Do you still think we should only do this `vroom()` for now or should we tackle the whole package? FWIW I've found 6 other locations where vroom inherits docs from readr.

jennybc: Note to selves: this still doesn't mention a really important feature which is that `vroom()` works with multiple files or connections. But I assume that's going to get handled with other multi-file issues that are in flight (?).

jennybc (approved):
LGTM

PR #468, opened 2022-09-30
Pass cli bullets to lifecycle

lionel-:
`details` will now be formatted with cli which causes this message to be width-wrapped, losing the newlines. With a future of cli, this sort of pattern will be made easier with "raw bullets".

### DIFF
--- a/R/path.R
+++ b/R/path.R
@@ -47,14 +47,14 @@ standardise_path <- function(path) {
 
     if (any(grepl("\n", path))) {
       lifecycle::deprecate_soft("1.5.0", "vroom(file = 'must use `I()` for literal data')",
-        details = glue::glue('
-
-          # Bad:
-          vroom("foo\\nbar\\n")
-
-          # Good:
-          vroom(I("foo\\nbar\\n"))
-        ')
+        details = c(
+          "",
+          "# Bad:",
+          "vroom(\"foo\\nbar\\n\")",
+          "",
+          "# Good:",
+          "vroom(I(\"foo\\nbar\\n\"))"
+        )
       )
       return(list(chr_to_file(path, envir = parent.frame())))
     }
--- a/R/path.R
+++ b/R/path.R
@@ -21,7 +21,7 @@ reencode_file <- function(path, encoding) {
 }
 
 # These functions adapted from https://github.com/tidyverse/readr/blob/192cb1ca5c445e359f153d2259391e6d324fd0a2/R/source.R
-standardise_path <- function(path, user_env = caller_env(2)) {
+standardise_path <- function(path, user_env = rlang::caller_env(2)) {
   if (is.raw(path)) {
     return(list(rawConnection(path, "rb")))
   }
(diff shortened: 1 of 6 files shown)


lionel-:
The second commit passes the user environment to `deprecate_soft()` so that it can determine whether usage was direct (in which case it will warn) or indirect (in which case it won't). Currently all usages were deemed indirect except in testthat.

Direct usage is determined as follows:

- The user env inherits from the global env.
- In testthat, the user env inherits from the namespace of the package being tested.

R/path.R
```diff
@@ -21,7 +21,7 @@ reencode_file <- function(path, encoding) {
 }
 
 # These functions adapted from https://github.com/tidyverse/readr/blob/192cb1ca5c445e359f153d2259391e6d324fd0a2/R/source.R
-standardise_path <- function(path) {
+standardise_path <- function(path, user_env = caller_env(2)) {
```
jennybc, line 24: What's the origin of this `2`? What I'm wondering is ... does this effectively hard-wire the way `standardise_path()` is currently used? I.e. how many calls are between the user and the call to `standardise_path()`?

lionel-: It's the same `caller_env(2)` taken by `deprecate_soft()` (and now also `deprecate_warn()`. You can also pass the `user_env` directly if you prefer.

The direct `caller_env()` of `deprecate_soft()` is the package that implements the deprecated feature (we pluck the e.g. namespace name from there). The user `caller_env(2)` of `deprecate_soft()` is the user of the deprecated feature. In case of soft-deprecation, we only warn if the user env is from global (direct usage in scripts) or is from a package currently being tested (direct usage in a package).

tests/testthat/test-path.R
```diff
@@ -61,6 +61,17 @@ test_that("split_path_ext works", {
...
 
+test_that("vroom() informs user to use I() for literal data (or not)", {
+  expect_snapshot(
+    x <- vroom("a,b,c,d\n1,2,3,4", show_col_types = FALSE)
+  )
+
+  f <- function(file) vroom(file, show_col_types = FALSE)
+  expect_snapshot(
+    x <- f("a,b,c,d\n1,2,3,4")
+  )
```
jennybc, line 72: I expected this scenario to _not_ warn the user 🤔

jennybc (commented):
I'm taking this chance to add a snapshot test. Shouldn't arm's-length usage lack the warning here?

tests/testthat/test-path.R
lionel-: Because it's called from a function? But that function inherits either from global (if you run the test interactively) or from your package namespace (if you run the test through `test()` [load-all'd namespace] or `check()` [cloned namespace]). So it counts as a direct usage.

If you want to reassure yourself of the indirect usage, you'll have to construct a function that inherits from another namespace like rlang or utils. We do this in the lifecycle tests but I wouldn't expect a package to do such tests. It's a bit tricky because you can't rely on the lexical scope of the unit test and instead ns-qualify everything and pass all data by argument or by injection. For this simple function though, that's only a matter of adding:

```r
environment(f) <- ns_env("utils")
```

Then this will count as an indirect usage because the warning will be emitted from a package that is not being tested.

PR #478, opened 2022-11-09
Update col_types.R

### LINKED ISSUE #477: Printing structure using `str` fails with `col_big_integer`
The following command fails in rstudio when colorizing the output of `str` because `col_big_integer` is not in the switch statement:
```
vroom(I("x,y\n1,2\n3,4\n"), col_types = list(y=col_integer(), x=col_big_integer())) |> str()
```
Note this does work when the output is not colorized for example in `reprex`, I believe the `col_big_integer` is omitted here:
https://github.com/tidyverse/vroom/blob/548344f4826818ca988e80b90cdf7fd837a90561/R/col_types.R#L253

bart1:
fix #477 prevent printing error on structure with col_big_integer

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* `str()` now works in a colourised context in the presence of a column of class `integer64`, i.e. parsed with `col_big_integer()` (@bart1, #477).
+
 # vroom 1.6.0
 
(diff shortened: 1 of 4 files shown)


jennybc:
Thanks!

jennybc (approved):

PR #480, opened 2022-12-07
Replace sprintf() with snprintf()

jeroen:
In `vroom_write.cc` we know the size of the output buffer (defined in the line above) so we use that to limit the size.

The two cases in `grisu3.c` we don't know the size of the buffer, so we limit to the input size+1. (+1 for the null terminator).


jennybc:
I'm going to merge this, so I can start revdep and other pre-release checks. But I think I probably will revisit this and try to make the file's provenance and evolution more clear. Thanks again for the patches.

PR #513, opened 2023-09-24
Remove unused code and variables

### LINKED ISSUE #504: Compilation warning about unused variable
```
In file included from vroom_fwf.cc:7:
./columns.h:125:10: warning: variable 'to_parse' set but not used [-Wunused-but-set-variable]
  size_t to_parse = 0;
```

https://github.com/tidyverse/vroom/blob/89b6aac8c9c07981a52c28251578deca11810ebd/src/columns.h#L125C10-L131

Looks real? i.e. `to_parse` doesn't actually get used beyond this loop

jennybc:
Closes #504

### DIFF
--- a/src/columns.h
+++ b/src/columns.h
@@ -92,7 +92,6 @@ inline cpp11::list create_columns(
     size_t num_threads) {
 
   R_xlen_t num_cols = idx->num_columns();
-  auto num_rows = idx->num_rows();
 
   auto locale_info = std::make_shared<LocaleInfo>(locale);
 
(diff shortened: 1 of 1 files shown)


src/columns.h
```diff
@@ -92,7 +92,6 @@ inline cpp11::list create_columns(
     size_t num_threads) {
 
   R_xlen_t num_cols = idx->num_columns();
-  auto num_rows = idx->num_rows();
```
jennybc, line 95: This line and those removed below date back to 04cc4fc.

`num_rows` is only used in the lines removed below, so it becomes an unused variable and generates a compilation warning once we remove those lines.

src/columns.h
```diff
@@ -122,14 +121,6 @@ inline cpp11::list create_columns(
...
       altrep);
 
-  size_t to_parse = 0;
-  for (R_xlen_t col = 0; col < num_cols; ++col) {
-    auto collector = my_collectors[col];
-    if (collector.use_altrep()) {
-      to_parse += num_rows;
-    }
-  }
-
```
jennybc, line 132: When this block was introduced in ??, it was followed by this line:

```
// Rcpp::Rcerr << to_parse << '\n';
```

which makes me think perhaps it primarily existed for print debugging / development purposes. Seems to strengthen the case for deletion vs. trying to figure out if there is some unfinished plan here.

PR #514, opened 2023-09-26
Allow reading list of connections

### LINKED ISSUE #509: `vroom` does not work on list of connections
I was following this example from the vignette but it seems `vroom` does not accept a list of connection I get the following message

```r
> vroom::vroom(list(file("file1.csv"), file("file2.csv")))
Error in enc2utf8(path) : argument is not a character vector
``` 

https://github.com/tidyverse/vroom/blob/89b6aac8c9c07981a52c28251578deca11810ebd/vignettes/vroom.Rmd#L113-L118

bairdj:
Fixes #509 

This amends `standardise_path` to work with a list of connections, as referred to in the vignette.

The test for this functionality "vroom works with many connections" was incorrectly passing a character vector of paths to `vroom` instead of a list of connections, so this wasn't being picked up as a failing test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* It is now possible (again?) to read from a list of connections (@bairdj, #514).
+
 * Internal change for compatibility with cpp11 >= 0.4.6 (@DavisVaughan).
 
 # vroom 1.6.3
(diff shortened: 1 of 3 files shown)


jennybc:
Thanks! I'm going to merge this, then make some refinements after the fact FYI.

> The test for this functionality "vroom works with many connections" was incorrectly passing a character vector of paths to vroom instead of a list of connections, so this wasn't being picked up as a failing test.

It's not very apparent, but reading from `*.csv.gz` files actually hits the connections code path. So the pre-existing test does actually do what it claims, in a way.

Issue #532, opened 2024-03-25
Feature request: Column-level na values via collectors


khusmann:
Right now, `NA` values are specified globally via `na` arg in the `read_*()` family of functions. Sometimes I want to supply `NA` values for specific *columns*, rather than the entire data set. A nice way to do this could be to add an `na` arg to all of the collector types to specify column-level missing values.

Column-level missing values come up frequently in survey data. Here are two examples:

Example 1:

> What is your current stress level?
> a. Low (LOW)
> b. Moderate (MODERATE)
> c. High (HIGH)
> d. I don’t know (DONT_KNOW)
> e. I don’t understand the question (DONT_UNDERSTAND)

I'd like to be able to create a `col_factor` type that reads the last two responses as `NA` as follows:

```
col_factor(levels = c("LOW", "MODERATE", "HIGH"), ordered = TRUE, na = c("DONT_KNOW", "DONT_UNDERSTAND"))
```

Example 2:

An item that records the individual's height as a double, but can have the following missing values: "ABSENT", "RULER_BROKE"

```
col_double(na = c("ABSENT", "RULER_BROKE"))
```

PR #541, opened 2024-07-18
Add collector-level na support

khusmann:
This PR adds support for collector-level `na` args (#532). This way, different lists of missing values can be specified for each column, overriding the global `na` arg in the call to `vroom()`.

Example:

``` r
vroom(
  I("a,b,c\na,foo,REFUSED\nb,REFUSED,MISSING\nOMITTED,bar,OMITTED\n"),
  col_types = cols(
    a = col_character(na = "OMITTED"),
    b = col_character(na = "REFUSED"),
    c = col_character()
  ),
  na = "MISSING"
)
#> # A tibble: 3 × 3
#>   a     b     c      
#>   <chr> <chr> <chr>  
#> 1 a     foo   REFUSED
#> 2 b     NA    NA     
#> 3 NA    bar   OMITTED
```

Without this PR, it is very difficult to efficiently read columns with different lists of missing values. Instead, they have to be loaded as character vectors, then parsed with `readr::parse_*()` or `readr::type_convert()`. There are two problems with this:

- parsing a chr vector after loading with vroom forces the vector to materialize, defeating vroom's lazy-loading altrep goodness
- vroom & readr's parsing rules have slightly diverged in subtle ways (e.g. https://github.com/tidyverse/readr/issues/1526)

I'm hoping you'll consider this PR for inclusion to vroom – it only requires a few changes, is 100% backwards compatible, and adds a feature that cannot otherwise be implemented in a separate package (without duplicating all of vroom's internals). Please let me know if there is anything more I can do to advocate for it. Thank you for your consideration!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -79,4 +79,4 @@ Copyright: file COPYRIGHTS
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3.9000
+RoxygenNote: 7.3.2
--- a/R/col_types.R
+++ b/R/col_types.R
@@ -32,6 +32,11 @@
 #' @param .delim The delimiter to use when parsing. If the `delim` argument
 #'   used in the call to `vroom()` it takes precedence over the one specified in
 #'   `col_types`.
+#' @param na Character vector of strings to interpret as missing values for the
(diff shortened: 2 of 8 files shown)


khusmann:
Note that this is failing the check for windows-latest (3.6) because the runner is grabbing the latest version of evaluate, which now requires R >= 4.0.0.

Issue #561, opened 2025-11-14
C API compliance


jennybc:
Current status of vroom:

```
Version: 1.6.6
Check: compiled code
Result: WARN 
  File ‘vroom/libs/vroom.so’:
    Found non-API calls to R: ‘DATAPTR’, ‘SETLENGTH’, ‘SET_TRUELENGTH’,
      ‘STDVEC_DATAPTR’
  This entry point may be removed soon:
  ‘STDVEC_DATAPTR’
```

I doubt I will tackle `SETLENGTH` and `SET_TRUELENGTH` at this time. I will definitely address `STDVEC_DATAPTR` and probably/presumably `DATAPTR`.

The email from CRAN doesn't explicitly distinguish between the urgency of these, so the situation is a bit unclear/fluid.

> Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_vroom.html>.
>
> Specifically, please see the WARNING in the r-devel compiled code check
about using non-API entry points which may be removed soon.  These have
replacements documented in "Some API replacements for non-API entry
points" of "Writing R Extensions": please change to use these
replacements.
>
>
> Please correct before 2025-11-27 to safely retain your package on CRAN.

jennybc:
#562 dealt with `STDVEC_DATAPTR()` and `DATAPTR()`, which is likely all I will do before a release. Larger forces are at work on the `SETLENGTH`/`SET_TRUELENGTH` front and I'm going to sit tight and wait for the dust to settle before choosing a course of action. I suspect they will continue to be tolerated with a NOTE for a while yet.

PR #562, opened 2025-11-17
Get rid of STDVEC_DATAPTR() and DATAPTR() usage

jennybc:
Related to #561

### DIFF
--- a/src/altrep.cc
+++ b/src/altrep.cc
@@ -98,7 +98,7 @@ bool vroom_altrep(SEXP x) {
       case LGLSXP: {
         SET_VECTOR_ELT(out, col, Rf_allocVector(LGLSXP, nrow));
         int* out_p = LOGICAL(VECTOR_ELT(out, col));
-        int* in_p = LOGICAL(elt);
+        const int* in_p = LOGICAL_RO(elt);
         for (R_xlen_t row = 0; row < nrow; ++row) {
           out_p[row] = in_p[row];
         }
@@ -107,7 +107,7 @@ bool vroom_altrep(SEXP x) {
       case INTSXP: {
         SET_VECTOR_ELT(out, col, Rf_allocVector(INTSXP, nrow));
         int* out_p = INTEGER(VECTOR_ELT(out, col));
-        int* in_p = INTEGER(elt);
+        const int* in_p = INTEGER_RO(elt);
         for (R_xlen_t row = 0; row < nrow; ++row) {
           out_p[row] = in_p[row];
         }
@@ -116,7 +116,7 @@ bool vroom_altrep(SEXP x) {
       case REALSXP: {
         SET_VECTOR_ELT(out, col, Rf_allocVector(REALSXP, nrow));
         double* out_p = REAL(VECTOR_ELT(out, col));
-        double* in_p = REAL(elt);
+        const double* in_p = REAL_RO(elt);
         for (R_xlen_t row = 0; row < nrow; ++row) {
           out_p[row] = in_p[row];
         }
@@ -125,9 +125,9 @@ bool vroom_altrep(SEXP x) {
       case STRSXP: {
         SET_VECTOR_ELT(out, col, Rf_allocVector(STRSXP, nrow));
         SEXP out_elt = VECTOR_ELT(out, col);
-        force_materialization(elt);
+        const SEXP* in_p = STRING_PTR_RO(elt);
         for (R_xlen_t row = 0; row < nrow; ++row) {
-          SET_STRING_ELT(out_elt, row, STRING_ELT(elt, row));
+          SET_STRING_ELT(out_elt, row, in_p[row]);
         }
         break;
       }


src/altrep.h
```diff
@@ -39,4 +39,10 @@ extern "C" {
 #include <R_ext/Altrep.h>
 }
 #endif
+
+// Backport DATAPTR_RW for R < 4.6.0 (as recommended in Writing R Extensions)
+#if R_VERSION < R_Version(4, 6, 0)
+# define DATAPTR_RW(x) DATAPTR(x)
+#endif
+
```
jennybc, line 47: Using guidance I see here: https://github.com/wch/r-source/commit/4a47d53a4a9c81b19a2e21e42549525a42d02d41

.github/workflows/R-CMD-check.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: any::rcmdcheck
+          extra-packages: any::rcmdcheck, patchwork@1.3.1
```
jennybc, line 65: Unrelated to the actual matter at hand, but necessary to pass R CMD check on Windows on oldrel-4.

jennybc:
Re: moving away from `DATAPTR()`, this feels like a real nothingburger, i.e. I can't really tell what has been accomplished. This ends up routing through `DATAPTR()`, one way or another. But it makes the warning go away, so yay?

src/altrep.cc
```diff
@@ -16,7 +16,7 @@
 
 [[cpp11::register]] void force_materialization(SEXP x) {
 #ifdef HAS_ALTREP
-  DATAPTR(x);
+  DATAPTR_RW(x);
```
DavisVaughan: If you're following this advice
https://github.com/wch/r-source/commit/2369435a1104b2c50ff92bd824244da6261a7703

then this is incorrect usage of `DATAPTR_RW()`.

We should really be calling something like the underlying `Materialize()` method for each type directly. Calling `DATAPTR()` ends up calling `Materialize()` as a side benefit.

src/vroom_big_int.h
```diff
@@ -99,7 +99,7 @@ class vroom_big_int : public vroom_vec {
   }
 
   static void* Dataptr(SEXP vec, Rboolean) {
-    return STDVEC_DATAPTR(Materialize(vec));
+    return REAL(Materialize(vec));
   }
```
DavisVaughan, line 103: These `Dataptr` methods _are_ actually the one place we are supposed to use `DATAPTR_RW()`

(I think we should consistently use `DATAPTR_RW()` in all `Dataptr` methods)

src/vroom_rle.h
```diff
@@ -96,15 +96,15 @@ class vroom_rle {
...
   }
 
   static const void* Dataptr_or_null(SEXP vec) {
     SEXP data2 = R_altrep_data2(vec);
     if (data2 == R_NilValue)
       return nullptr;
 
-    return STDVEC_DATAPTR(data2);
+    return DATAPTR_RO(data2);
   }
```
DavisVaughan, line 108: This looks correct to me 👍

jennybc:
I took the 2 concrete suggestions. One wholesale, the other with a tweak.

* "Use DATAPTR_RW() in all Dataptr methods, not REAL() and friends". Sure, done. I ended up with `REAL()` and friends somewhat due to the order in which I tackled this. Removing `STDVEC_DATAPTR()` was job one. And I swear the WRE advice changed while I was looking into this? As in, I think the `DATAPTR_RW()` recommendation may have appeared during this process.
* "`force_materialization()` and `vroom_materialize()` and `vroom_convert()` should really do something other than call `DATAPTR_RW()` to "force materialization by side effect".... we need some way to call our per-type `Materialize()` methods directly." I've inlined this in the existing `force_materialization()` and then called that in more places. Adding `vroom_altrep_materialize()` just felt like a possibly unnecessary new layer of indirection.

src/altrep.cc
```diff
@@ -40,15 +59,13 @@ bool vroom_altrep(SEXP x) {
 [[cpp11::register]] SEXP vroom_materialize(SEXP x, bool replace) {
 #ifdef HAS_ALTREP
   for (R_xlen_t col = 0; col < Rf_xlength(x); ++col) {
-
     SEXP elt = VECTOR_ELT(x, col);
-    // First materialize all of the non-character vectors
```
jennybc, line 45: I think this mention of "non-character vectors" was leftover from a version where there were two branches. Removal felt better than correction.

src/altrep.cc
```diff
@@ -108,7 +125,7 @@ bool vroom_altrep(SEXP x) {
       case STRSXP: {
         SET_VECTOR_ELT(out, col, Rf_allocVector(STRSXP, nrow));
         SEXP out_elt = VECTOR_ELT(out, col);
-        DATAPTR(elt);
+        force_materialization(elt);
```
jennybc: We're still discussing the fact that this will only do anything for a vroom-originating character altrep column.

DavisVaughan: On closer investigation it looks like you could actually use `STRING_PTR_RO()` here, making the string path more symmetric with the other integer/double/logical paths, and removing the problem entirely

```cpp
      case STRSXP: {
        SET_VECTOR_ELT(out, col, Rf_allocVector(STRSXP, nrow));
        SEXP out_elt = VECTOR_ELT(out, col);
        const SEXP* in_p = STRING_PTR_RO(elt);
        for (R_xlen_t row = 0; row < nrow; ++row) {
          SET_STRING_ELT(out_elt, row, in_p[row]);
        }
        break;
      }
```

Under the hood, `STRING_PTR_RO()` will call the ALTREP `Dataptr` method with `writable = false`, forcing materialization

That would be my suggestion now

DavisVaughan: (Really the other paths should use the "RO" version of the accessors as well, like `double* in_p = REAL(elt);` should really be `const double* in_p = REAL_RO(elt);` to emphasize that it is "Read-Only" access to `elt`, giving you a `const double*` not a `double*`)

jennybc:
@DavisVaughan Take another look? I think this might be done. I'm pretty sure the sole CI failure is due to today's stormy weather on the internet (affects tests trying to real from a URL). I'm seeing the same failures atm locally.

Issue #554, opened 2025-04-23
vroom::problems() conflicts and gives misleading warning when readr is loaded?


MatthieuStigler:
When both `vroom` and `readr` are loaded, using the bare  `vroom::problems()` will give a misleading error message:

> Error in `vroom::problems()`:
#> ! The `x` argument of `vroom::problems()` must be a data frame created
#>   by vroom:
#> ✖ `x` seems to have been created with something else, maybe readr? 

``` r
library(vroom)
library(readr)
#> Registered S3 methods overwritten by 'readr':
#>   method                    from 
#>   as.data.frame.spec_tbl_df vroom
#>   as_tibble.spec_tbl_df     vroom
#>   format.col_spec           vroom
#>   print.col_spec            vroom
#>   print.collector           vroom
#>   print.date_names          vroom
#>   print.locale              vroom
#>   str.col_spec              vroom
#> 
#> Attaching package: 'readr'
#> The following objects are masked from 'package:vroom':
#> 
#>     as.col_spec, col_character, col_date, col_datetime, col_double,
#>     col_factor, col_guess, col_integer, col_logical, col_number,
#>     col_skip, col_time, cols, cols_condense, cols_only, date_names,
#>     date_names_lang, date_names_langs, default_locale, fwf_cols,
#>     fwf_empty, fwf_positions, fwf_widths, locale, output_column,
#>     problems, spec


input_file <- vroom_example("mtcars.csv")
v <- vroom(input_file)
#> Rows: 32 Columns: 12
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr  (1): model
#> dbl (11): mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
vroom::problems(v)
#> # A tibble: 0 × 5
#> # ℹ 5 variables: row <int>, col <int>, expected <chr>, actual <chr>, file <chr>

v <- vroom(input_file, col_select = c(model, cyl, gear))
#> Rows: 32 Columns: 3
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr (1): model
#> dbl (2): cyl, gear
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
[...]

PR #565, opened 2025-11-21
Guard against loss of class

jennybc:
Related to #554

### DIFF
--- a/tests/testthat/test-vroom.R
+++ b/tests/testthat/test-vroom.R
@@ -1307,3 +1307,50 @@ test_that("vroom can read a datetime column with no data and skip 1", {
     equals = tibble::tibble(dt = as.POSIXct(character()))
   )
 })
+
+# https://github.com/tidyverse/vroom/issues/554
+# https://github.com/tidyverse/vroom/issues/534
+test_that("vroom(col_select =) output has 'spec_tbl_df' class, spec, and problems when readr is attached", {
+  # Register readr's [.spec_tbl_df method which drops attributes and the "spec_tbl_df" class
+  readr_single_bracket <- function(x, ...) {
+    attr(x, "spec") <- NULL
+    attr(x, "problems") <- NULL
+    class(x) <- setdiff(class(x), "spec_tbl_df")
+    NextMethod(`[`)
+  }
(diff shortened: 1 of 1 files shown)


R/col_types.R
```diff
@@ -374,12 +374,7 @@ vroom_enquo <- function(x) {
 }
 
 vroom_select <- function(x, col_select, id) {
-  spec <- attr(x, "spec")
```
jennybc, line 377: The need to store and restore attributes comes from the possibility that a call to `x[whatever]` could drop attributes, depending on the S3 dispatch for `[`. If we need to preserve `spec` then we also need preserve `problems`, i.e. they should get the same treatment, whatever that may be.

In the current solution, `x` won't have any "special" subclasses yet and `[` should preserve attributes.

R/col_types.R
```diff
@@ -374,12 +374,7 @@ vroom_enquo <- function(x) {
...
-
-  # Drop any NULL columns
-  is_null <- vapply(x, is.null, logical(1))
-  x[is_null] <- NULL
-  # reorder and rename columns
```
jennybc, line 382: This was already being done in the body of `vroom()` and done intentionally before the call to `tibble::as_tibble()`. Doing it here also is redundant and confusing.

R/vroom.R
```diff
@@ -296,9 +296,9 @@ vroom <- function(
   }
 
   out <- tibble::as_tibble(out, .name_repair = identity)
-  class(out) <- c("spec_tbl_df", class(out))
```
jennybc, line 299: Swapping the order of `vroom_select()` and applying the `"spec_tbl_df"` class is the main fix here.

R/vroom_fwf.R
```diff
@@ -102,6 +102,10 @@ vroom_fwf <- function(
     progress = progress
   )
 
+  # Drop any NULL columns
+  is_null <- vapply(out, is.null, logical(1))
+  out[is_null] <- NULL
+
```
jennybc, line 108: It was actually a bug that this wasn't here, before the call to `tibble::as_tibble()`. It just hasn't been reported, I think.

Issue #303, opened 2021-02-03
vroom::spec returns NULL when col_select is specified and readr is loaded


kent37:
When `readr` is loaded, calling `vroom::vroom` with a `col_select` argument returns a data frame without a `spec` attribute so `vroom::spec()` returns `NULL`.

``` r
library(vroom)

# This works
df <- vroom(vroom_example("mtcars.csv"), col_select=-mpg)
#> Rows: 32
#> Columns: 11
#> Delimiter: ","
#> chr [ 1]: model
#> dbl [10]: cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb
#> 
#> Use `spec()` to retrieve the guessed column specification
#> Pass a specification to the `col_types` argument to quiet this message

vroom::spec(df)
#> cols(
#>   model = col_character(),
#>   mpg = col_skip(),
#>   cyl = col_double(),
#>   disp = col_double(),
#>   hp = col_double(),
#>   drat = col_double(),
#>   wt = col_double(),
#>   qsec = col_double(),
#>   vs = col_double(),
#>   am = col_double(),
#>   gear = col_double(),
#>   carb = col_double(),
#>   .delim = ","
#> )

library(readr)
#> Registered S3 methods overwritten by 'readr':
#>   method           from 
#>   format.col_spec  vroom
#>   print.col_spec   vroom
#>   print.collector  vroom
#>   print.date_names vroom
#>   print.locale     vroom
#>   str.col_spec     vroom
#> 
#> Attaching package: 'readr'
#> The following objects are masked from 'package:vroom':
#> 
#>     as.col_spec, col_character, col_date, col_datetime, col_double,
#>     col_factor, col_guess, col_integer, col_logical, col_number,
#>     col_skip, col_time, cols, cols_condense, cols_only, date_names,
#>     date_names_lang, date_names_langs, default_locale, fwf_cols,
#>     fwf_empty, fwf_positions, fwf_widths, locale, output_column,
#>     problems, spec

# This returns a data frame with no 'spec' attribute
df <- vroom(vroom_example("mtcars.csv"), col_select=-mpg)

vroom::spec(df)
#> NULL
```

<sup>Created on 2021-02-03 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value                       
[...]

PR #567, opened 2025-11-23
Guard against loss of class

### LINKED ISSUE #534: `problems()` does not get set when readr is loaded, *and* `col_select` is being used.
(Running vroom 1.6.5 & readr 2.1.5 on Ubuntu)

When readr is loaded in the environment, and `col_select` option is being used, the `problems()` attr gets no longer get attached to the resulting tibble:

``` r
library(readr)

bar <- vroom::vroom(
  I("a\n1\n2\nz\n"),
  delim = ",",
  col_types=cols(a = col_double()),
  col_select="a"
)

problems(bar) # No output! attr(bar, "problems") == NULL!
```

Interestingly, when `col_select` is not used, it works just fine:

``` r
library(readr)

bar <- vroom::vroom(
  I("a\n1\n2\nz\n"),
  delim = ",",
  col_types=cols(a = col_double())
)

problems(bar)
#> e.g.:
#>   dat <- vroom(...)
#>   problems(dat)
#> # A tibble: 1 × 5
#>     row   col expected actual file                              
#>   <int> <int> <chr>    <chr>  <chr>                             
#> 1     4     1 a double z      /tmp/Rtmp62LnNu/file1d276c1662f970
```

It also works just fine when readr is not loaded:

``` r
library(vroom)

bar <- vroom(
  I("a\n1\n2\nz\n"),
  delim = ",",
  col_types=cols(a = col_double()),
  col_select="a"
)

problems(bar)
#> e.g.:
#>   dat <- vroom(...)
#>   problems(dat)
#> # A tibble: 1 × 5
#>     row   col expected actual file                             
#>   <int> <int> <chr>    <chr>  <chr>                            
#> 1     4     1 a double z      /tmp/RtmpYwCnXv/file1d26632e67a7f
```

### LINKED ISSUE #554: vroom::problems() conflicts and gives misleading warning when readr is loaded?

jennybc:
Fixes #554, fixes #534, retroactively provides an alternative fix for #303 (#566 cleared the way for nicer fixes here)

The problem is that `vroom()` was applying the `"spec_tbl_df"` class before calling `vroom_select()`, which uses `[` internally. If readr is attached, this dispatches readr's `[.spec_tbl_df` method, which is designed to drop the `"spec_tbl_df"` class and the `spec` and `problems` attributes. The target use case is for downstream user manipulation of a tibble that was originally parsed by readr or vroom. But it's basically a mistake to use this `[.spec_tbl_df` while you're still in the process of creating the tibble.

(Side note: the order of operations was correct in `vroom_fwf()` already.)

One thing this has brought up for me is ... why *doesn't* vroom have a `[.spec_tbl_df` method of its own? 🤔 Seems like the same considerations would apply, which is that, once a user starts to subset a vroom-produced tibble, you'd want to ditch `spec` and `problems` and the subclass (the original motiavtion is spelled out in https://github.com/tidyverse/readr/commit/e94b8bc99c172d7c5b7a4877b3503d3ff01265d5). vroom conditionally registers practically every other S3 method that readr does. Why not this one? Recorded this question in #569.

### DIFF
--- a/tests/testthat/test-vroom.R
+++ b/tests/testthat/test-vroom.R
@@ -1307,3 +1307,50 @@ test_that("vroom can read a datetime column with no data and skip 1", {
     equals = tibble::tibble(dt = as.POSIXct(character()))
   )
 })
+
+# https://github.com/tidyverse/vroom/issues/554
+# https://github.com/tidyverse/vroom/issues/534
+test_that("vroom(col_select =) output has 'spec_tbl_df' class, spec, and problems when readr is attached", {
+  # Register readr's [.spec_tbl_df method which drops attributes and the "spec_tbl_df" class
+  readr_single_bracket <- function(x, ...) {
+    attr(x, "spec") <- NULL
+    attr(x, "problems") <- NULL
+    class(x) <- setdiff(class(x), "spec_tbl_df")
+    NextMethod(`[`)
+  }
+  registerS3method("[", "spec_tbl_df", readr_single_bracket)
+
+  # Unfortunately you can't just de-register the method and
+  # local_mocked_s3_method() only works if method exists to begin with.
+  # We'll do next best thing which is to put a pass-through method back.
+  withr::defer({
+    registerS3method("[", "spec_tbl_df", function(x, ...) NextMethod(`[`))
+  })
+
+  expect_warning(
+    dat <- vroom(
+      I("a,b\n1,2\nz,3\n4,5"),
+      col_types = "dc",
+      col_select = c(a, b),
+      show_col_types = FALSE,
+      altrep = FALSE
+    ),
+    class = "vroom_parse_issue"
+  )
+
+  expect_s3_class(dat, "spec_tbl_df")
+  expect_equal(
+    attr(dat, "spec", exact = TRUE),
+    cols(
+      a = col_double(),
+      b = col_character(),
+      .delim = ","
+    )
+  )
+  expect_no_error(probs <- problems(dat))
+  if (exists("probs")) {
+    expect_equal(nrow(probs), 1)
+    expect_equal(probs$row, 3)
+    expect_equal(probs$col, 1)
+  }
+})
--- a/tests/testthat/test-vroom_fwf.R
+++ b/tests/testthat/test-vroom_fwf.R
@@ -468,3 +468,51 @@ test_that("vroom_fwf correctly reads DOS files with no trailing newline (https:/
 
   expect_equal(out, out2)
 })
+
+# https://github.com/tidyverse/vroom/issues/554
+# https://github.com/tidyverse/vroom/issues/534
+test_that("vroom_fwf(col_select =) output has 'spec_tbl_df' class, spec, and problems when readr is attached", {
+  # Register readr's [.spec_tbl_df method which drops attributes and the "spec_tbl_df" class
+  readr_single_bracket <- function(x, ...) {
+    attr(x, "spec") <- NULL
+    attr(x, "problems") <- NULL
+    class(x) <- setdiff(class(x), "spec_tbl_df")
+    NextMethod(`[`)
+  }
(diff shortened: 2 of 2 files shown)


R/col_types.R
```diff
@@ -374,8 +374,6 @@ vroom_enquo <- function(x) {
 }
 
 vroom_select <- function(x, col_select, id) {
-  spec <- attr(x, "spec")
```
jennybc, line 377: The need to store and restore attributes comes from the possibility that a call to `x[whatever]` could drop attributes, depending on the S3 dispatch for `[`. If we need to preserve `spec` then we also need preserve `problems`, i.e. they should get the same treatment, whatever that may be.

In the current solution, `x` won't have any "special" subclasses yet and `[` should preserve attributes.

R/vroom.R
```diff
@@ -292,9 +292,9 @@ vroom <- function(
   }
 
   out <- tibble::as_tibble(out, .name_repair = identity)
-  class(out) <- c("spec_tbl_df", class(out))
```
jennybc, line 295: Applying `"spec_tbl_df"` class *after* calling `vroom_select()` is the main fix here.

jennybc:
The CI failure on macOS is due to some weirdness with the runner, I think. I've been seeing it intermittently for the past couple of days and seems to come from a lockfile being read too soon after it's been written.

PR #571, opened 2025-11-26
rchk fixes

### LINKED ISSUE #444: Consider rchk issues
https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/vroom.out

```
Package vroom version 1.5.7
Package built using 82396/R 4.3.0; x86_64-pc-linux-gnu; 2022-05-24 14:34:41 UTC; unix   
Checked with rchk version 35d5d01a98903a5f3e9ec166dcdb2f790ecb784f LLVM version 12.0.1
More information at https://github.com/kalibera/cran-checks/blob/master/rchk/PROTECT.md
For rchk in docker image see https://github.com/kalibera/rchk/blob/master/doc/DOCKER.md

Suspicious call (two or more unprotected arguments) to Rf_lang3 at vroom_errors::warn_for_errors() const vroom/src/./vroom_errors.h:89

Function read_chr(vroom_vec_info*)::$_1::operator()() const
  [UP] unprotected variable val while calling allocating function vroom::iterator::filename() const vroom/src/vroom_chr.cc:31
  [UP] unprotected variable val while calling allocating function vroom::iterator::index() const vroom/src/vroom_chr.cc:31
  [UP] unprotected variable val while calling allocating function cpp11::writable::r_vector<cpp11::r_string>::operator SEXPREC*() const vroom/src/vroom_chr.cc:34

Function vroom_errors::warn_for_errors() const
  [UP] calling allocating function Rf_findFun(S:warn,?) with argument allocated using Rf_findVarInFrame(?,S:rlang) vroom/src/./vroom_errors.h:86
```

jennybc:
Closes #444

On the basis of https://github.com/tidyverse/vroom/actions/runs/19692713991/job/56411718420, I believe that this fixes the 2 rchk results I am actually trying to address.

### DIFF
--- a/src/vroom_errors.h
+++ b/src/vroom_errors.h
@@ -90,7 +90,7 @@ public:
       SEXP cli_ns = Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("cli"));
       PROTECT(cli_ns);
       SEXP cli_warn = Rf_findFun(Rf_install("cli_warn"), cli_ns);
-      UNPROTECT(1);
+      PROTECT(cli_warn);
       cpp11::strings bullets({
         "w"_nm = "One or more parsing issues, call {.fun problems} on your data frame for details, e.g.:",
         " "_nm = "dat <- vroom(...)",
@@ -100,6 +100,7 @@ public:
         bullets,
         Rf_mkString("vroom_parse_issue"));
       Rf_eval(cli_warn_call, R_EmptyEnv);
+      UNPROTECT(2);
     }
   }
 


src/vroom_errors.h
```diff
@@ -87,9 +87,10 @@ class vroom_errors {
       have_warned_ = true;
       // it is intentional that we aren't using cpp11::package
       // https://github.com/tidyverse/vroom/commit/984a3e5e37e124feacfec3d184dbeb02eb1145c4
-      static auto cli_warn = Rf_findFun(
-          Rf_install("cli_warn"),
-          Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("cli")));
+      SEXP cli_ns = Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("cli"));
+      PROTECT(cli_ns);
+      SEXP cli_warn = Rf_findFun(Rf_install("cli_warn"), cli_ns);
+      PROTECT(cli_warn);
```
jennybc, line 93: I realize this means we might look up `cli::cli_warn()` more than once per R session (which is how I interpret the `static auto` of the original). But I still think the lookup will occur at most once per call to a high-level vroom call, which seems fine to me. We only warn once for a given data frame (see lines 86 and 87 above).

PR #576, opened 2025-12-03
Switch from `stop()` to `cli::cli_abort()`

### LINKED ISSUE #413: Update errors to rlang 1.0.0. Helpful guides:https://rlang.r-lib.org/reference/topic-error-call.htmlhttps://rlang.r-lib.org/reference/topic-error-chaining.htmlhttps://rlang.r-lib.org/reference/topic-condition-formatting.html

jennybc:
Closes #413

### DIFF
--- a/R/vroom_fwf.R
+++ b/R/vroom_fwf.R
@@ -124,7 +124,7 @@ fwf_empty <- function(
   file <- standardise_one_path(standardise_path(file)[[1]])
 
   if (inherits(file, "connection")) {
-    stop("`file` must be a regular file, not a connection", call. = FALSE)
+    cli::cli_abort("{.arg file} must be a regular file, not a connection.")
   }
 
   if (n < 0 || is.infinite(n)) {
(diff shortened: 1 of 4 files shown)


R/vroom_fwf.R
```diff
@@ -167,25 +167,45 @@ fwf_positions <- function(start, end = NULL, col_names = NULL) {
...
-#'   variable widths if a length one vector, and if length two, variable start and end
-#'   positions. The elements of `...` are used to construct a data frame
-#'   with or or two rows as above.
+#' @param ... Named or unnamed arguments, each addressing one column. Each input
+#' should be either a single integer (a column width) or a pair of integers
+#' (column start and end positions). All arguments must have the same shape,
+#' i.e. all widths or all positions.
 fwf_cols <- function(...) {
   x <- lapply(list(...), as.integer)
+
```
jennybc, line 176: I actually cleaned up the logic in this function, updated the docs to match what actually happens (previously they suggested you could provide a data frame, but that seems to have never been true), and improved/updated the error messages.

R/vroom_fwf.R
```diff
@@ -199,11 +219,12 @@ fwf_col_names <- function(nm, n) {
 verify_fwf_positions <- function(col_positions) {
   is_greater <- stats::na.omit(col_positions$begin > col_positions$end)
   if (any(is_greater)) {
-    bad <- which(is_greater)
-    stop(
-      "`col_positions` must have begin less than end.\n* Invalid values at position(s): ",
```
jennybc, line 204: This wording was not technically correct, so changed it.

CLAUDE.md
```diff
@@ -15,8 +15,8 @@ General advice:
 ### Testing
 
 - Tests for `R/{name}.R` go in `tests/testthat/test-{name}.R`.
-- Use `devtools::test(reporter = "check")` to run all tests
-- Use `devtools::test(filter = "name", reporter = "check")` to run tests for `R/{name}.R`
+- Use `devtools::test()` to run all tests
+- Use `devtools::test(filter = "name")` to run tests for `R/{name}.R`
```
jennybc, line 19: `reporter = "check"` is tickling this new feature in testthat:

> New `extract_test()` function to extract a reprex from a failing expectation. tests run in `R CMD check` will use this to automatically create a reprex in the `_problems/` directory for each failing expectation. You can turn this behaviour off by setting TESTTHAT_PROBLEMS=false (#2263).

Experimenting with _not_ requesting the check reporter.

PR #582, opened 2025-12-05
Phase in the new resizable vector API

### LINKED ISSUE #561: C API compliance

jennybc:
Closes #561

This is the last move to get into C API compliance 🎉

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,7 +21,7 @@
   e.g., a nonexistent URL (#488).
 
 * vroom takes the recommended approach for phasing out usage of the non-API
-  entry points `SETLENGTH`, `SET_GROWABLE_BIT`, and `SET_TRUELENGTH` (#582).
+  entry points `SETLENGTH` and `SET_TRUELENGTH` (#582).
 
 # vroom 1.6.7
 


jennybc:
@DavisVaughan I don't think there's anything that interesting to see here, just recruiting a fellow human.

src/altrep.h
```diff
@@ -46,3 +46,11 @@ extern "C" {
 #endif
 
 #endif
+
+// Backport resizable vector API for R < 4.6.0
+// See: https://rstudio.github.io/r-manuals/r-exts/The-R-API.html#moving-into-c-api-compliance
+#if R_VERSION < R_Version(4, 6, 0)
+# define R_allocResizableVector(type, maxlen) Rf_allocVector(type, maxlen)
+# define R_duplicateAsResizable(x) Rf_duplicate(x)
+# define R_resizeVector(x, newlen) SETLENGTH(x, newlen)
```
DavisVaughan: check https://rstudio.github.io/r-manuals/r-exts/The-R-API.html#some-backports again, as it has updated since

src/columns.h
```diff
@@ -99,9 +100,11 @@ inline cpp11::list create_columns(
 
   bool add_filename = !Rf_isNull(id);
 
-  cpp11::writable::list res(num_cols + add_filename);
+  cpp11::writable::list res(
+      R_allocResizableVector(VECSXP, num_cols + add_filename));
 
-  cpp11::writable::strings res_nms(num_cols + add_filename);
+  cpp11::writable::strings res_nms(
+      R_allocResizableVector(STRSXP, num_cols + add_filename));
```
DavisVaughan, line 127: I am honestly quite nervous about what putting these in a `cpp11::writable::list` or `cpp11::writable::strings` will do. It...probably...works fine, but this is so experimental it is hard to know.

Would it be possible to put them in a `cpp11::sexp` instead? I would feel more comfortable with that even though it will require more code changes below, because you'll need `SET_VECTOR_ELT()` and friends.

PR #588, opened 2025-12-12
Check for write errors when buffering connections to temp files

### LINKED ISSUE #544: segfault error reading large compressed file
Hi,

I cannot read a large zip file with vroom (v. 1.6.5) but I can read it once unzipped.

I saw there are closed similar issues, but none of the solutions seem to work for me and they are also supposed to have been fixed in the version I installed (if I understand correctly).

```
r$> download.file("https://s3.amazonaws.com/data.patentsview.org/download/g_inventor_disambiguated.tsv.zip")

r$> df <- "g_inventor_disambiguated.tsv.zip" |>
  vroom::vroom(
    col_select = c(inventor_id),
    col_types = vroom::cols(inventor_id = vroom::col_character())
  )

r$> df

 *** caught segfault ***
address 0x7f4dafee9009, cause 'memory not mapped'

Traceback:
 1: vec_slice(x, seq_len(n))
 2: vec_head(as.data.frame(x), n)
 3: df_head(x, n)
 4: tbl_format_setup.tbl(x, width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines, focus = focus)
 5: tbl_format_setup_dispatch(x, width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines, focus = focus)
 6: tbl_format_setup(x, width = width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines, focus = attr(x, "pillar_focus"))
 7: format_tbl(x, width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines)
 8: format.tbl(x, width = width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines)
 9: format(x, width = width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines)
10: writeLines(format(x, width = width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines))
11: print_tbl(x, width, ..., n = n, max_extra_cols = max_extra_cols,     max_footer_lines = max_footer_lines)
12: print.tbl(x)
13: (function (x, ...) UseMethod("print"))(x)

Possible actions:
1: abort (with core dump, if enabled)
2: normal R exit
3: exit R without saving workspace
4: exit R saving workspace
```

Here some information about my environment
```
r$> sessionInfo()
R version 4.3.1 (2023-06-16)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: Debian GNU/Linux 12 (bookworm)

Matrix products: default
BLAS/LAPACK: /home/******/miniforge3/envs/******/lib/libopenblasp-r0.3.27.so;  LAPACK version 3.12.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8
 [8] LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

time zone: Europe/Rome
tzcode source: system (glibc)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
[1] compiler_4.3.1
```

Please let me know if you need any extra information.
I hope this helps to solve the issue.

jennybc:
Fixes #544

In ☝️ I confirm that I can reproduce the problem, which demonstrates that it's a "ran out of space" problem. I won't repeat that here.

General orientation to this bug:

Reading from `.zip`, `.gz`, `.bz2`, or `.xz` files (or any R connection, including URLs) requires different handling inside vroom than regular files. Regular files are memory-mapped directly from disk. But data from compressed files or connections must be written to a temp file first.

vroom does this work in chunks (default 128 KB buffer): read a chunk from the connection, write it to the temp file, and index it - all with async operations running in parallel. The temp file persists for the lifetime of the vroom data frame, as it's what back the altrep vectors and is cleaned up by R's garbage collector.

The bug I'm fixing here: During this chunked writing+indexing process, if an individual `fwrite()` call fails partway through the process (e.g., disk full after 1 GB of a 2 GB file), we don't clue in, because we aren't checking the return value. The async write task fails silently, but vroom soldiers on, reading and indexing all the way to the end. This leaves us with an index built from the full data stream (the whole ~2 GB in the example), but the memory map only covers the partial file that was actually written (1 GB). Later, when we try to materialize values using positions from the index that are beyond the memory-mapped region, we segfault.

The fix: Now we're constantly checking the `fwrite()` return value (in the async lambda) and `ferror()` on the `FILE` stream before memory-mapping (sort of a belt-and-suspenders situation, I think). If any write fails, we delete the partial temp file and provide a clear error message with the temp directory path and approximate file size, suggesting either freeing space or setting `VROOM_TEMP_PATH`.

I did wonder ... as soon as we see a failed write, why don't we stop reading and indexing? But (1) this process really doesn't take much time, (2) this is a rare edge case, and we're in an async multi-threaded context ... I'm happy to let that initial task complete (and fail), then surface the problem before we move on to memory mapping.

@DavisVaughan I feel pretty good about the diagnosis and fix. So I'm just looking for a second set of eyes and any questions that come to mind. That always seems to lead to better understanding and/or code.

### DIFF
--- a/src/delimited_index_connection.cc
+++ b/src/delimited_index_connection.cc
@@ -276,7 +276,7 @@ delimited_index_connection::delimited_index_connection(
     }
 
     ss << "To fix this:\n"
-       << "  * Free up disk space in your temporary directory\n"
+       << "  * Free up disk space in your temporary directory.\n"
        << "  * Or set VROOM_TEMP_PATH to a directory with more space:\n"
        << "    `Sys.setenv(\"VROOM_TEMP_PATH\" = \"/path/to/larger/disk\")`";
 
--- a/src/fixed_width_index_connection.cc
+++ b/src/fixed_width_index_connection.cc
@@ -191,7 +191,7 @@ fixed_width_index_connection::fixed_width_index_connection(
     }
 
     ss << "To fix this:\n"
-       << "  * Free up disk space in your temporary directory\n"
+       << "  * Free up disk space in your temporary directory.\n"
        << "  * Or set VROOM_TEMP_PATH to a directory with more space:\n"
        << "    `Sys.setenv(\"VROOM_TEMP_PATH\" = \"/path/to/larger/disk\")`";
 


src/delimited_index_connection.cc
```diff
@@ -209,8 +214,13 @@ delimited_index_connection::delimited_index_connection(
         break;
       }
     }
-    write_fut = std::async(
-        [&, i, sz] { std::fwrite(buf[i].data(), sizeof(char), sz, out); });
+    total_bytes_to_write += sz;
```
jennybc: It feels nice to give the user a sense of how much space appears to be needed.

src/delimited_index_connection.cc
```diff
@@ -209,8 +214,13 @@ delimited_index_connection::delimited_index_connection(
...
+    write_fut = std::async([&, i, sz] {
+      size_t written = std::fwrite(buf[i].data(), sizeof(char), sz, out);
+      if (written != sz) {
+        write_error.store(true);
```
jennybc, line 219: This is the heart of the matter: noticing when we are not able to complete the intended write.

src/delimited_index_connection.cc
```diff
@@ -235,15 +245,48 @@ delimited_index_connection::delimited_index_connection(
...
+      double mb = total_bytes_to_write / (1024.0 * 1024.0);
+      ss << "Bytes attempted to write: ~" << static_cast<int>(mb) << " MB\n\n";
+    }
+
+    ss << "To fix this:\n"
+       << "  * Free up disk space in your temporary directory\n"
+       << "  * Or set VROOM_TEMP_PATH to a directory with more space:\n"
+       << "    `Sys.setenv(\"VROOM_TEMP_PATH\" = \"/path/to/larger/disk\")`";
+
+    cpp11::stop("%s", ss.str().c_str());
```
jennybc, line 283: Here's how this looks when you run afoul of it now:

```
Error:
! Failed to write temporary file when reading from connection.
This usually means there is not enough disk space.

Temporary directory: /Volumes/VroomTest
Bytes attempted to write: ~2 GB

To fix this:
  * Free up disk space in your temporary directory
  * Or set VROOM_TEMP_PATH to a directory with more space:
    `Sys.setenv("VROOM_TEMP_PATH" = "/path/to/larger/disk")`
```

Image

src/fixed_width_index_connection.cc
```diff
@@ -6,8 +6,10 @@
 #include "r_utils.h"
 #include "unicode_fopen.h"
 #include <array>
+#include <atomic>
```
jennybc, line 9: Basically applying the same patch as for `src/delimited_index_connection.cc` and moving a few statements around to remove incidental differences between `src/delimited_index_connection.cc` and `src/fixed_width_index_connection.cc`.

src/delimited_index_connection.cc
DavisVaughan: Is this different from `total_read`?

src/delimited_index_connection.cc
```diff
@@ -235,15 +245,48 @@ delimited_index_connection::delimited_index_connection(
   if (write_fut.valid()) {
     write_fut.wait();
   }
+
+  if (should_close) {
+    cpp11::package("base")["close"](in);
+  }
+
+  // Check for write errors before closing
```
DavisVaughan, line 251: This comment seems wrong. You close before checking for write errors.

src/delimited_index_connection.cc
```diff
@@ -235,15 +245,48 @@ delimited_index_connection::delimited_index_connection(
...
+  bool had_write_error = write_error.load() || std::ferror(out);
```
DavisVaughan, line 252: When would `std::ferror(out)` be needed? `write_error` feels like enough to me.

src/delimited_index_connection.cc
jennybc: No, it describes that you have to capture the error status on `out` before you can `fclose()` the file stream. Now, that begs the question of why we're checking that ...

src/delimited_index_connection.cc
jennybc: I asked this exact question when working on the PR. But claude convinced me that, while the information should be redundant in theory, it's a standard C pattern for file I/O to do this sort of belt-and-suspenders thing. And it seems harmless.

src/delimited_index_connection.cc
jennybc: It is not, now that I look at `total_read`. Removing `total_bytes_to_write` here and on the fixed width side, in favor of using `total_read`.

PR #595, opened 2026-01-06
Remove arrangements for older versions of R that vroom does not support

### LINKED ISSUE #594: Remove guards and workarounds form R < 4.1
vroom already has a hard requirement for R >= 4.1:

https://github.com/tidyverse/vroom/blame/main/DESCRIPTION#L29

The more I frolic around the C++ code, the more bummed out I am by version guards and arrangements that are no longer necessary. I'm going to rationalize this.

jennybc:
Closes #594

@DavisVaughan I'm asking for review just to get a second set of eyeballs on this and make sure I'm not missing any obvious downside. I feel like this is a net positive for the codebase.

More context re: why I want to clean up dusty bits: I have a medium term goal of moving the guts of vroom into readr (readr really _is_ vroom at this point, for the vast majority of usage) and thereby eliminating lots of annoyance around maintenance, documentation, error messaging, name collisions, etc. At the same time, standalone vroom would be put into a frozen state and on a path to eventual archival.


R/vroom.R
```diff
@@ -526,16 +524,16 @@ vroom_altrep <- function(which = NULL) {
...
-    getRversion() >= "3.5.0" && which$dttm %||% vroom_use_altrep_dttm(),
-    getRversion() >= "3.5.0" && which$date %||% vroom_use_altrep_date(),
-    getRversion() >= "3.5.0" && which$time %||% vroom_use_altrep_time(),
-    getRversion() >= "3.5.0" && which$big_int %||% vroom_use_altrep_big_int()
+    which$chr %||% vroom_use_altrep_chr(),
+    which$fct %||% vroom_use_altrep_fct(),
+    which$int %||% vroom_use_altrep_int(),
+    which$dbl %||% vroom_use_altrep_dbl(),
+    which$num %||% vroom_use_altrep_num(),
+    which$lgl %||% vroom_use_altrep_lgl(),
```
jennybc, line 532: vroom actually does not ever use altrep for logical columns, but I gather there must have been some thought that it might one day? Anyway, I'm not getting into that here.

src/columns.h
```diff
@@ -168,53 +151,39 @@ inline cpp11::list create_columns(
...
         SET_VECTOR_ELT(res, i, read_num(info));
         delete info;
       }
       break;
     case column_type::Lgl:
-      // if (collector.use_altrep()) {
-      //#if defined HAS_ALTREP && R_VERSION >= R_Version(3, 6, 0)
-      // SET_VECTOR_ELT(res, i, vroom_lgl::Make(info));
-      //#endif
-      //} else {
```
jennybc, line 214: More evidence of vague-but-unrealized plans around altrep logical columns.

README.md
```diff
@@ -236,15 +234,6 @@ There are also individual variables for each type. Currently only
...
 
-## RStudio caveats
-
-RStudio’s environment pane calls `object.size()` when it refreshes the
-pane, which for Altrep objects can be extremely slow. RStudio 1.2.1335+
-includes the fixes
-([RStudio#4210](https://github.com/rstudio/rstudio/pull/4210),
-[RStudio#4292](https://github.com/rstudio/rstudio/pull/4292)) for this
-issue, so it is recommended you use at least that version.
-
```
jennybc, line 247: This now feels like ancient history that need not be mentioned.

R/col_types.R
```diff
@@ -624,15 +624,11 @@ collector_value.collector_factor <- function(x, ...) {
   factor()
 }
 
-# the more obvious as.POSIXct(double()) doesn't work on R < 4.0
-# https://github.com/tidyverse/vroom/issues/453
```
jennybc, line 628: I like the vctrs implementation and don't feel like it needs justification.

src/connection.h
```diff
@@ -10,43 +10,6 @@
 #undef FALSE
 #include <cpp11/logicals.hpp>
 
-#ifdef VROOM_USE_CONNECTIONS_API
```
jennybc, line 13: The connections API never happened (or un-happened, I guess), so doing some cleanup around that too.

R/col_types.R
DavisVaughan: we also have `vctrs::new_datetime()` since base R doesn't really provide good constructors

R/vroom.R
DavisVaughan: Probably the fact that it didn't exist in 3.5 delayed its implementation, and then it was never added later on?

jennybc: Yeah? But I also think the performance argument is so different (less compelling) for logical. I'd at least want to look into that quantitatively before ever considering *adding* that feature.

PR #596, opened 2026-01-07
Use r altrep class getters

### LINKED ISSUE #593: Get rid of ATTRIB
`ATTRIB` has recently appeared in the CRAN checks for r-devel (I've already eliminated `SETLENGTH` and `SET_TRUELENGTH` in dev vroom):

```
Result: NOTE 
  File ‘vroom/libs/vroom.so’:
    Found non-API calls to R: ‘ATTRIB’, ‘SETLENGTH’, ‘SET_TRUELENGTH’
```

I'm not under a new deadline yet, but this makes me suspect that, when it happens, it will also include `ATTRIB`.

The usage is here:

https://github.com/tidyverse/vroom/blob/d56888edeff033e234df54a89da34f89ef56133d/src/altrep.cc#L151-L152

`ATTRIB` joined the non-API list in this commit: https://github.com/wch/r-source/commit/c711f19109224acbbed2a63f19e78d5eaf5a311d

This commit suggests that vroom should switch to the new `R_altrep_class_name()` and `R_altrep_class_package()`: https://github.com/wch/r-source/commit/37eb29515a83672b53743cba11e88167ca063d06

jennybc:
Closes #593

### DIFF
--- a/src/altrep.h
+++ b/src/altrep.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <cpp11/R.hpp>
 
 #include <R_ext/Rdynload.h>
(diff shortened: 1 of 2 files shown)


src/altrep.h
```diff
@@ -1,48 +1,26 @@
+#pragma once
```
jennybc, line 1: This should have really been here before, but it becomes absolutely necessary now that we've got some inline function definitions here.

PR #599, opened 2026-01-11
Download remote compressed files

### LINKED ISSUE #400: vroom doesn't respect skip/comment options for remote (HTTP) gzipped CSVs
I've recently hit a strange edge case where `vroom` doesn't respect `skip` (nor `comment`) options when reading a gziped CSV on a remote URL connection. Reprex is below that shows an empty tibble being created. Downloading the file does allow the read to work as expected, as does operating on a non-GZiped remote file.

I've seen some issues on `readr` that suggest a limitation with the base R connection readers that may account for this, but I'm unable to see any known limitations that say this shouldn't work. Hoping this may be a fix that can be implemented here (and then picked up in readr).

Please let me know if there is any additional info I can give for debugging. Thanks for the great package!

``` r
library(vroom)
url <- "https://epss.cyentia.com/epss_scores-2022-02-04.csv.gz"

#this fails
vroom(url, comment ="#")
#> New names:
#> * `` -> ...1
#> Rows: 0 Columns: 1
#> ── Column specification ────────────────────────────────────────────────────────
#> 
#> chr (1): ...1
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> # A tibble: 0 × 1
#> # … with 1 variable: ...1 <chr>

#this works
tmp <- tempfile()
download.file(url, tmp)
vroom(tmp, comment = "#")
#> Rows: 168325 Columns: 3
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr (1): cve
#> dbl (2): epss, percentile
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> # A tibble: 168,325 × 3
#>    cve              epss percentile
#>    <chr>           <dbl>      <dbl>
#>  1 CVE-2021-42013 0.816       0.995
#>  2 CVE-2021-1732  0.0910      0.869
#>  3 CVE-2021-4034  0.0364      0.666
#>  4 CVE-2013-1763  0.0382      0.673
#>  5 CVE-2014-3153  0.0230      0.583
#>  6 CVE-2019-17497 0.0525      0.759
#>  7 CVE-2018-4993  0.722       0.991
#>  8 CVE-2021-44228 0.944       0.999
#>  9 CVE-2019-5420  0.931       0.999
#> 10 CVE-2021-3156  0.587       0.986
#> # … with 168,315 more rows
unlink(tmp)
```

<sup>Created on 2022-02-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessionInfo()
#> R version 4.1.2 (2021-11-01)
#> Platform: x86_64-apple-darwin17.0 (64-bit)
#> Running under: macOS Big Sur 10.16
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] vroom_1.5.7
#> 
#> loaded via a namespace (and not attached):
#>  [1] pillar_1.6.2      compiler_4.1.2    highr_0.9         R.methodsS3_1.8.1
#>  [5] R.utils_2.11.0    tools_4.1.2       digest_0.6.27     bit_4.0.4        
#>  [9] evaluate_0.14     lifecycle_1.0.0   tibble_3.1.6      R.cache_0.15.0   
#> [13] pkgconfig_2.0.3   rlang_1.0.0       reprex_2.0.1      cli_3.0.1        
#> [17] rstudioapi_0.13   curl_4.3.2        parallel_4.1.2    yaml_2.2.1       
#> [21] xfun_0.29         fastmap_1.1.0     withr_2.4.2       styler_1.6.2     
#> [25] stringr_1.4.0     knitr_1.37        fs_1.5.0          vctrs_0.3.8      
#> [29] bit64_4.0.5       tidyselect_1.1.1  glue_1.6.1        fansi_0.5.0      
#> [33] rmarkdown_2.10    tzdb_0.2.0        purrr_0.3.4       magrittr_2.0.1   
#> [37] backports_1.2.1   ellipsis_0.3.2    htmltools_0.5.2   utf8_1.2.2       
#> [41] stringi_1.7.4     crayon_1.4.1      R.oo_1.24.0
```

</details>

### LINKED ISSUE #553: incomplete download from URL
Motived by https://stackoverflow.com/questions/79587028/read-csv-only-reads-a-fraction-of-rows-from-a-zipped-file-when-reading-from-url

A gzipped file is only being downloaded partially, whereas reading in the pre-downloaded file reads in every row. Notice `429497` rows when loaded as a local file, `1798` when loaded by URL.

```r
download.file("https://www1.ncdc.noaa.gov/pub/data/ghcn/daily/by_station/USW00014839.csv.gz", "USW00014839.csv.gz")
# trying URL 'https://www1.ncdc.noaa.gov/pub/data/ghcn/daily/by_station/USW00014839.csv.gz'
# Content type 'application/gzip' length 1699132 bytes (1.6 MB)
# ==================================================
# downloaded 1.6 MB
vroom::vroom("USW00014839.csv.gz")
# New names:
# • `` -> `...5`
# • `` -> `...6`
# • `` -> `...8`
# Warning: One or more parsing issues, call `problems()` on your data frame for details, e.g.:
#   dat <- vroom(...)
#   problems(dat)
# Rows: 429497 Columns: 8
# ── Column specification ─────────────────────────────────────────────────────────────────────────────────────────────
# Delimiter: ","
# chr (4): USW00014839, TMAX, ...5, X
# dbl (3): 19380401, 22, ...8
# lgl (1): ...6
# ℹ Use `spec()` to retrieve the full column specification for this data.
# ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
# # A tibble: 429,497 × 8
#    USW00014839 `19380401` TMAX   `22` ...5  ...6  X      ...8
#    <chr>            <dbl> <chr> <dbl> <chr> <lgl> <chr> <dbl>
#  1 USW00014839   19380402 TMAX      6 <NA>  NA    X        NA
#  2 USW00014839   19380403 TMAX     33 <NA>  NA    X        NA
#  3 USW00014839   19380404 TMAX     33 <NA>  NA    X        NA
#  4 USW00014839   19380405 TMAX      0 <NA>  NA    X        NA
#  5 USW00014839   19380406 TMAX      0 <NA>  NA    X        NA
#  6 USW00014839   19380407 TMAX     17 <NA>  NA    X        NA
#  7 USW00014839   19380408 TMAX     22 <NA>  NA    X        NA
#  8 USW00014839   19380409 TMAX     56 <NA>  NA    X        NA
#  9 USW00014839   19380410 TMAX    117 <NA>  NA    X        NA
# 10 USW00014839   19380411 TMAX    178 <NA>  NA    X        NA
# # ℹ 429,487 more rows
# # ℹ Use `print(n = ...)` to see more rows
vroom::vroom("https://www1.ncdc.noaa.gov/pub/data/ghcn/daily/by_station/USW00014839.csv.gz", col_names = FALSE)
# Warning: One or more parsing issues, call `problems()` on your data frame for details, e.g.:
#   dat <- vroom(...)
#   problems(dat)
# Rows: 1798 Columns: 8
# ── Column specification ─────────────────────────────────────────────────────────────────────────────────────────────
# Delimiter: ","
# chr (3): X1, X3, X7
# dbl (2): X2, X4
# lgl (3): X5, X6, X8
# ℹ Use `spec()` to retrieve the full column specification for this data.
# ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
# # A tibble: 1,798 × 8
#    X1                X2 X3       X4 X5    X6    X7    X8   
#    <chr>          <dbl> <chr> <dbl> <lgl> <lgl> <chr> <lgl>
#  1 USW00014839 19380401 TMAX     22 NA    NA    X     NA   
#  2 USW00014839 19380402 TMAX      6 NA    NA    X     NA   
#  3 USW00014839 19380403 TMAX     33 NA    NA    X     NA   
#  4 USW00014839 19380404 TMAX     33 NA    NA    X     NA   
#  5 USW00014839 19380405 TMAX      0 NA    NA    X     NA   
#  6 USW00014839 19380406 TMAX      0 NA    NA    X     NA   
#  7 USW00014839 19380407 TMAX     17 NA    NA    X     NA   
#  8 USW00014839 19380408 TMAX     22 NA    NA    X     NA   
#  9 USW00014839 19380409 TMAX     56 NA    NA    X     NA   
# 10 USW00014839 19380410 TMAX    117 NA    NA    X     NA   
# # ℹ 1,788 more rows
# # ℹ Use `print(n = ...)` to see more rows
```

The `problems()` suggestion does not help much, it's mostly about a few parsing errors indicating an incorrect guess at `lgl`:

```r
# # A tibble: 3 × 5
#     row   col expected           actual file 
#   <int> <int> <chr>              <chr>  <chr>
# 1    15     6 1/0/T/F/TRUE/FALSE I      ""   
# 2    44     6 1/0/T/F/TRUE/FALSE I      ""   
# 3   858     6 1/0/T/F/TRUE/FALSE S      ""   
```

I tested at least one other gzipped file on that site (`USW00014838.csv.gz`), and it shows the same behavior, albeit at a different cut point (6628 of 204830 rows). It does not appear to be a clear networking issue, as both `download.file` (shown above) and `data.table::fread` (not shown) read in the entire file (both, actually) without message/warning/error. 

<details><summary> <code>session_info()</code> </summary>

```r
sessioninfo::session_info(pkgs = "vroom")
# ─ Session info ────────────────────────────────────────────────────────────────────────────────────────────────────
#  setting  value
#  version  R version 4.4.3 (2025-02-28)
#  os       macOS Sequoia 15.4.1
#  system   aarch64, darwin20
#  ui       X11
#  language (EN)
#  collate  en_US.UTF-8
#  ctype    en_US.UTF-8
#  tz       America/New_York
#  date     2025-04-22
#  pandoc   3.6.4 @ /opt/homebrew/opt/pandoc/bin/ (via rmarkdown)
#  quarto   1.6.43 @ /usr/local/bin/quarto
# ─ Packages ────────────────────────────────────────────────────────────────────────────────────────────────────────
#  package     * version date (UTC) lib source
#  bit           4.5.0.1 2024-12-03 [1] RSPM (R 4.4.0)
#  bit64         4.6.0-1 2025-01-16 [1] RSPM (R 4.4.0)
#  cli           3.6.4   2025-02-13 [1] RSPM (R 4.4.3)
#  cpp11         0.5.1   2024-12-04 [1] RSPM (R 4.4.0)
#  crayon        1.5.3   2024-06-20 [1] RSPM (R 4.4.3)
#  fansi         1.0.6   2023-12-08 [1] RSPM (R 4.4.3)
#  glue          1.8.0   2024-09-30 [1] RSPM (R 4.4.3)
#  hms           1.1.3   2023-03-21 [1] RSPM (R 4.4.0)
#  lifecycle     1.0.4   2023-11-07 [1] RSPM (R 4.4.3)
#  magrittr      2.0.3   2022-03-30 [1] RSPM (R 4.4.3)
#  pillar        1.10.1  2025-01-07 [1] RSPM (R 4.4.0)
#  pkgconfig     2.0.3   2019-09-22 [1] RSPM (R 4.4.3)
#  prettyunits   1.2.0   2023-09-24 [1] RSPM (R 4.4.3)
#  progress      1.2.3   2023-12-06 [1] RSPM (R 4.4.0)
#  R6            2.6.0   2025-02-12 [1] RSPM (R 4.4.0)
#  rlang         1.1.5   2025-01-17 [1] RSPM (R 4.4.0)
#  tibble        3.2.1   2023-03-20 [1] RSPM (R 4.4.3)
#  tidyselect    1.2.1   2024-03-11 [1] RSPM (R 4.4.3)
#  tzdb          0.4.0   2023-05-12 [1] RSPM (R 4.4.0)
#  utf8          1.2.4   2023-10-22 [1] RSPM (R 4.4.3)
#  vctrs         0.6.5   2023-12-01 [1] RSPM (R 4.4.3)
#  vroom         1.6.5   2023-12-05 [1] RSPM (R 4.4.0)
#  withr         3.0.2   2024-10-28 [1] RSPM (R 4.4.3)
#  [1] /Users/r2/Library/R/4.4
#  [2] /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library
# ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
</details>

(Other users on SO reported the same behavior on this link. I'm posting here as a courtesy.)

jennybc:
Fixes #400
Fixes #553
Fixes https://github.com/tidyverse/readr/issues/1555
Fixes https://github.com/tidyverse/readr/issues/1553 (speculative, but there's no reprex)

Download remote compressed files (`.gz`, `.bz2`, `.xz`, `.zip`) to a temp file, then treat them like a local compressed file.

* For `.bz2`, `.xz` and `.zip`, this is a new capability. Previously this errored with "not supported, download locally first".
* For `.gz`, this is a change of approach. vroom already did support remote `.gz`, but it optimistically used `base::gzcon()` to avoid downloading the compressed file. Problem is that `gzcon()` has a bug (https://bugs.r-project.org/show_bug.cgi?id=18887) that was briefly fixed, but the fix got reverted. Remote `.gz` files that are concatenated gzip archives (multiple gzip members per [RFC 1952](https://www.rfc-editor.org/rfc/rfc1952.html)) silently return truncated data. `gzcon()` only reads the first gzip member (boo), whereas `gzfile()` reads all of them (yay).

It's easy to convince yourself that the unsuccessful `vroom()` call is due to *something else* (3 of the 4 issues above start out barking up the wrong tree, all different trees), but it's really about a concatenated `.gz` file.

---

From the Chesterton's Fence perspective ... I cannot find any evidence in the git history of any specific reason to _not_ download to a temp file. Nor can I (or Claude) think of some gotcha. I think it was just an accident of history and maybe the fact that, at least for `.gz`, you could stream compressed data with `gzcon()`. (Also, FWIW, data.table downloads the whole file in the background, then decompresses it.)

---

There was some question about where to put the new pre-download functionality. My first instinct was in `connection_or_filepath()`, but that is part of a call chain that starts in R, then goes to C++, then back to R. Given the current wiring, it's not possible to schedule cleanup of the temp file at the right time. I see how to fix that (something like https://github.com/jennybc/parentframecpp/pull/1), but I'm going to open a separate issue (#601) and PR for this, because there are several areas in vroom that would benefit from a similar approach.

`standardise_path()` is another reasonable location and it's more favorable now, in terms of cleanup. So that's where I introduce this.

### DIFF
--- a/tests/testthat/test-path.R
+++ b/tests/testthat/test-path.R
@@ -78,8 +78,7 @@ test_that("vroom works via https on concatenated gz file", {
   skip_on_cran()
 
   mt <- vroom(vroom_example("mtcars.csv"), col_types = list())
-  # FIXME: switch the branch to main or HEAD here
-  url <- "https://raw.githubusercontent.com/tidyverse/vroom/download-remote-compressed-files/inst/extdata/mtcars.csv.gz"
+  url <- "https://raw.githubusercontent.com/tidyverse/vroom/HEAD/inst/extdata/mtcars-concatenated.csv.gz"
   expect_equal(vroom(url, comment = "#", col_types = list()), mt)
 })
 


R/path.R
```diff
@@ -108,6 +111,37 @@ standardise_path <- function(
...
+  # Handle remote compressed files from R, before heading to C++, so we can
+  # associate cleanup with the desired environment.
+  is_remote_compressed <- vapply(
+    path,
+    function(p) {
+      is_url(p) && tolower(tools::file_ext(p)) %in% c("gz", "bz2", "xz", "zip")
+    },
+    logical(1)
+  )
+  if (all(is_remote_compressed)) {
```
jennybc, line 123: It's not *strictly necessary* to be so "all or nothing", but we have a similar mindset re: multiple inputs for, e.g., connections and it makes things less fiddly. I could entertain better handling of mixed inputs if real users complain.

R/path.R
```diff
@@ -108,6 +111,37 @@ standardise_path <- function(
...
+    return(lapply(path, function(p) {
+      ext <- tolower(tools::file_ext(p))
+      local_path <- download_file(p, ext, call = call)
+      withr::defer(unlink(local_path), envir = call)
```
jennybc, line 127: This is the `defer()` that we want to place before we ever head into C++, given the current design.

R/path.R
```diff
@@ -108,6 +111,37 @@ standardise_path <- function(
...
+      switch(
+        ext,
+        gz = gzfile(local_path),
```
jennybc, line 130: `.gz` just gets treated like the rest now. Bye bye `base::gzcon()`.

R/path.R
```diff
@@ -132,32 +166,20 @@ connection_or_filepath <- function(path, write = FALSE, call = caller_env()) {
   }
 
   if (is_url(path)) {
+    ext <- tolower(tools::file_ext(path))
```
jennybc, line 169: This is probably a better location for the call to `download_file()` but for now we need to take care of it earlier.

R/path.R
```diff
@@ -222,10 +244,10 @@ connection_or_filepath <- function(path, write = FALSE, call = caller_env()) {
 
   switch(
     compression,
-    gz = gzfile(path, ""),
-    bz2 = bzfile(path, ""),
-    xz = xzfile(path, ""),
-    zip = zipfile(path, ""),
```
jennybc, line 228: `""` is the default value for the second argument `open` anyway. (I had to make that true for `zipfile()`, which is ours, but also seemed logical for consistency.)

R/path.R
```diff
@@ -319,6 +341,33 @@ is_url <- function(path) {
   grepl("^((http|ftp)s?|sftp)://", path)
 }
 
+download_file <- function(url, ext, call = caller_env()) {
+  local_path <- vroom_tempfile(fileext = ext, pattern = "vroom-download-url-")
```
jennybc, line 345: vroom promises to use `VROOM_TEMP_PATH` for all such needs, which is the main point of `vroom_tempfile()`.

R/path.R
```diff
@@ -319,6 +341,33 @@ is_url <- function(path) {
...
+  show_progress <- vroom_progress()
```
jennybc, line 346: Empirically, it's pretty important to show progress, even if it's not styled like vroom progress. The real-world examples from the motivating issues can take time to download and otherwise you think things have just hung.

R/vroom.R
```diff
@@ -176,7 +177,7 @@
 #' vroom(I("a|b\n1.0|2.0\n"), delim = "|")
 #'
 #' # Read datasets across multiple files ---------------------------------------
-#' mtcars_by_cyl <- vroom_example(vroom_examples("mtcars-"))
+#' mtcars_by_cyl <- vroom_example(vroom_examples("mtcars-[468]"))
```
jennybc, line 180: This regex had to tighten up in order to *not* match the new test fixture, `mtcars-concatenated.csv.gz`.

tests/testthat/test-path.R
```diff
@@ -49,14 +49,38 @@ test_that("vroom works via https on gz file", {
...
+  url <- "https://raw.githubusercontent.com/tidyverse/vroom/main/inst/extdata/mtcars.csv.zip"
+  expect_equal(vroom(url, col_types = list()), mt)
+})
+
+# https://github.com/tidyverse/vroom/issues/400
+test_that("vroom works via https on concatenated gz file", {
+  skip_on_cran()
+
+  mt <- vroom(vroom_example("mtcars.csv"), col_types = list())
+  # FIXME: switch the branch to main or HEAD here
```
jennybc: I'll either do this right before merge (and fail last CI on PR) or right after merge (and fail first CI on `main`).

R/path.R
DavisVaughan: Is `envir = call` the right frame? Or is it the environment of `standardise_path()`?

`call` is (typically) one frame above that, right?

Or, or, wow, is the "one frame above" `standardise_path()` the place where you actually read in the file? So you're trying to say "please clean up this file later on, after ive fully read it in, and i promise the timing of that will all work out"

If that is the case, then this definitely deserves a comment because this feels like a big old footgun that could easily be triggered (like if you add a layer of indirection between the call to `standardise_path()` and where the file actually gets read in, then you might clean it up too soon)

I'd also consider any other approach that might better remove this footgun

R/path.R
```diff
@@ -319,6 +341,33 @@ is_url <- function(path) {
...
+    },
+    error = function(cnd) {
+      unlink(local_path)
+      cli::cli_abort(
+        c(
+          "Failed to download {.url {url}}.",
+          "x" = conditionMessage(cnd)
+        ),
+        parent = NA,
+        error = cnd,
```
DavisVaughan, line 362: I don't know what `error =` is?

R/path.R
jennybc: Cleanup is definitely what required the most thought and is related to what I say in the intro re: where to put this "download to temporary file and schedule cleanup" in the overview. This thinking lead to #601 and https://github.com/jennybc/parentframecpp/pull/1. There are multiple places in vroom where a better approach to passing the calling environment around will benefit cleanup and error messaging.

But, yes, what's here currently does the cleanup at the right time. It goes down like so:

```
vroom()
      └── standardise_path(call = caller_env())  # call = vroom()'s env
              └── download_file() + withr::defer(..., envir = call)
      └── vroom_()  # reads the file
      └── <vroom returns, cleanup runs>
```

So the cleanup runs when `vroom()` exits, which is after the file has been consumed by `vroom_()` (as far as `vroom_()` knows, this is just a normal non-remote compressed file).

This fits an existing pattern in `chr_to_file()` and `reencode_file()`. Inspired by this convo, I'm tweaking those sites to be even more consistent and then all of them will get another round of 👀 in #601.

DavisVaughan: I think if I were implementing this, knowing how annoyingly complex this is, I think I would pass through both `call` and another env called `env_cleanup` or `env_defer`. Even if it is the same environment as `call`, it feels like it would express the meaning better 

When I see `call`, I'm conditioned to think "this has to do with error messaging", so seeing it used for other things confuses me and smells off to me.

I'd also be happy if some other approach nukes the need for this entirely, but I also wouldn't be mad if this were the final solution, where `vroom()` itself just has this at the very top:

```r
vroom <- function(...) {
  # The frame we tie all temporary file cleanup to
  env_cleanup <- environment()
}
```

R/path.R
jennybc: My intent is to pass the original condition along as data, if someone really wanted to catch it and dig in. Inspired by examples seen in [Taking full ownership of a causal error](https://rlang.r-lib.org/reference/topic-error-chaining.html#taking-full-ownership-of-a-causal-error).

> When a low-level error is overtaken, it is good practice to store it in the high-level error object, so that it can be inspected for debugging purposes. In the snippet above, we stored it in the `error` field.

R/path.R
DavisVaughan: Another thing I'd probaby do is make `env_cleanup` a _required_ argument to each helper, rather than setting to `caller_env()`, because that is never going to be the correct environment and would lull me into a false sense of correctness

jennybc: These are great ideas and I'm going to add them as part of the mandate in #601.

PR #604, opened 2026-01-16
Better handling of mismatched quotes

### LINKED ISSUE #484: ‘vroom’ silently truncates CSV files with mismatched quotes
See [the blog post](https://json.blog/2023/01/21/delimited-files-are.html) and [the discussion it triggered](https://www.reddit.com/r/rstats/comments/10imhat/delimited_files_are_hell_a_comparison_of_methods/).

Reprex:

```r
> vroom::vroom('x.csv', show_col_types = FALSE)
# A tibble: 0 × 4
# … with 4 variables: A <chr>, B <chr>, C <chr>, D <chr>
# ℹ Use `colnames()` to see all variable names

> data.table::fread('x.csv')
     A   B                   C  D
1: foo bar "ba\nz,bat\n1,2,3,4 NA
Warning message:
In data.table::fread("x.csv") :
  Detected 4 column names but the data has 3 columns. Filling rows automatically. Set fill=TRUE explicitly to avoid this warning.

> cat(paste(readLines('x.csv'), collapse = '\n'))
A,B,C,D
foo,bar,"ba
z,bat
1,2,3,4
```

Of course CSV encompasses various formats but even so it’s not clear why ‘vroom’ thinks that the file is valid but has no data rows (despite having more than one line). I therefore guess this is unintentional (I can't think of a situation where silently dropping rows would be the expected behaviour).

From a user perspective, there are two likely scenarios:

1. a broken file which is missing the closing quotation mark;
2. a file without quotes, and `quote = ""` should have been passed.

(2) is a user error and can thus be ignored here (in fact, passing `quote = ""` leads to warnings on the above file, which is expected). (1) should ideally generate a warning or even an error.

jennybc:
Fixes #484
Fixes https://github.com/tidyverse/readr/issues/1539
Fixes https://github.com/tidyverse/readr/issues/1491

When `vroom()` encounters an unclosed quote, the row containing it and all subsequent rows are silently omitted.

``` r
vroom::vroom(
  I(c('A,B,C', 'd,e,"f', 'g,h,i')),
  show_col_types = FALSE
)
#> # A tibble: 0 × 3
#> # ℹ 3 variables: A <chr>, B <chr>, C <chr>
```

With this PR, `vroom()` now returns that last, malformed row and the parsing problem is surfaced via the usual mechanism, i.e.g. via `problems()`.

``` r
library(vroom)

dat <- vroom(
  I(c('A,B,C', 'd,e,"f', 'g,h,i')),
  show_col_types = FALSE
)
problems(dat)
#> e.g.:
#>   dat <- vroom(...)
#>   problems(dat)
#> # A tibble: 1 × 5
#>     row   col expected      actual      file                                    
#>   <int> <int> <chr>         <chr>       <chr>                                   
#> 1     2     3 closing quote end of file /private/tmp/RtmpDt9n8e/vroom-chr-to-fi…

dat
#> # A tibble: 1 × 3
#>   A     B     C           
#>   <chr> <chr> <chr>       
#> 1 d     e     "f\ng,h,i\n"
```

High-level description of the fix: When we're indexing the entire file or stream and we finish in the `QUOTED_FIELD` state, we now finalize the final, garbled row and register a parsing problem.

In a separate comment, I'll put a concrete example that I found helpful / documents what I've learned about this area of the code base. I'll also comment inline on the most important changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # vroom (development version)
 
+* Unclosed quotes (e.g., `a,b,"c` with no closing `"`) now trigger a warning, instead of silent data truncation. The affected row is also newly included in the returned data, which should facilitate troubleshooting (#484, https://github.com/tidyverse/readr/issues/1539, https://github.com/tidyverse/readr/issues/1491).
+
 * [vroom.tidyverse.org](https://vroom.tidyverse.org/) is the new home of
   vroom's website, catching up to the much earlier move (April 2022) of vroom's
   GitHub repository from the r-lib organization to the tidyverse. The motivation


src/delimited_index.cc
```diff
@@ -152,7 +152,7 @@ delimited_index::delimited_index(
     idx_ = std::vector<idx_t>(num_threads + 1);
 
     // Index the first row
-    size_t cols = 0;
+    size_t num_delims = 0;
```
jennybc, line 155: This is just a renaming. But this variable literally counts the number of delimiters we've encountered within a record and the code makes more sense to me this way. Certainly less fiddly than having to contemplate whether `cols` starts counting from 0 or 1.

src/delimited_index.cc
```diff
@@ -179,7 +179,7 @@ delimited_index::delimited_index(
 
     std::vector<std::future<void>> threads;
 
-    if (nmax_set) {
+    if (num_threads == 1) {
```
jennybc, line 182: This is a meaningful change.

Background: Multi-threaded indexing splits the file into chunks, with each chunk creating its own local `state` variable initialized to `RECORD_START`. State doesn't flow between chunks - it's discarded when each chunk finishes. If a chunk encounters a newline while in `QUOTED_FIELD`, it throws `newline_error`, triggering a retry with `num_threads = 1`.

The problem with the old condition: After catching `newline_error`, we set `num_threads = 1` and `goto start_indexing`. But `nmax_set` is still false (it was computed before `start_indexing`), so `if (nmax_set)` would be false. The retry would go through `parallel_for` with 1 thread, which still creates a local `state` variable that gets discarded - so we wouldn't be able to detect the unclosed quote at EOF.

The fix: `if (num_threads == 1)` ensures the retry takes the stateful path, where `state` flows through the entire file and persists after indexing completes. This allows us to detect when we finish in `QUOTED_FIELD` state.

We've always set `num_thread` to 1 when `nmax_set` is true:

https://github.com/tidyverse/vroom/blob/2104925c70ddfdf9c5903a1c2feebe4864617ba6/src/delimited_index.cc#L134-L137

so I actually think it was almost a bug that this was not `num_threads == 1` even before this PR.

src/delimited_index.cc
```diff
@@ -261,6 +261,20 @@ delimited_index::delimited_index(
       t.get();
     }
 
+    // If we finish indexing a file and we're in QUOTED_FIELD, warn about an
+    // unclosed quote. This only applies to single-threaded indexing, where
+    // state flows through the entire file. Multi-threaded indexing throws
+    // newline_error as soon as a newline is seen inside QUOTED_FIELD.
+    if (num_threads == 1 && state == QUOTED_FIELD) {
+      errors->add_parse_error(
+          file_size, num_delims, "closing quote", "end of file");
```
jennybc, line 270: Note that we're recording info about what sort of error this is ("expected a closing quote, but found EOF instead").

src/delimited_index.h
```diff
@@ -272,24 +272,27 @@ class delimited_index : public index,
...
+    while (num_delims > 0 && num_delims >= num_cols) {
+      destination.pop_back();
+      --num_delims;
+    }
+    // If we didn't find enough delimiters, fall back to recording the end of
+    // row as the field start position. These artificial fields will have zero
+    // length.
+    while (num_delims < num_cols - 1) {
+      destination.push_back(pos);
+      ++num_delims;
```
jennybc, line 295: This diff is a combination of:

* Refactoring the code in a way that makes more sense to me.
* Using the new "shape" for `add_parse_error()`.

src/delimited_index_connection.cc
```diff
@@ -246,6 +246,17 @@ delimited_index_connection::delimited_index_connection(
     write_fut.wait();
   }
 
+  // If we finish indexing a connection and we're in QUOTED_FIELD, warn about
```
jennybc, line 249: Same idea, but now for the connection side of the house. Connections are processed in chunks, but sequentially, i.e. as a stream.

src/vroom_errors.h
```diff
@@ -16,7 +16,10 @@ class vroom_errors {
   struct parse_error {
     size_t position;
     size_t column;
-    parse_error(size_t pos, size_t col) : position(pos), column(col) {}
+    std::string expected;
+    std::string actual;
+    parse_error(size_t pos, size_t col, std::string exp, std::string act)
+        : position(pos), column(col), expected(exp), actual(act) {}
```
jennybc, line 22: The `parse_error` struct previously only stored `position` and `column`, and `resolve_parse_errors()` computed the error message assuming all errors were column count mismatches ("expected N columns, actual M columns").

For unclosed quotes, we need a different message: "expected closing quote, actual end of file". So we changed `parse_error` to store the expected/actual strings directly, making it flexible enough for both column count errors and unclosed quote errors.

(Also: confusingly, we use `add_parse_error()` to record *indexing* problems. And we use `add_error()` to record problems during *parsing*. That is neither here not there, but something I keep noticing.)

tests/testthat/test-vroom.R
jennybc, line 1: According to codecov, the "patch" for this PR has 100% coverage.

src/delimited_index.cc
DavisVaughan: I think you are right, but I would also accept a stupidly explicit if statement of

```
if (num_threads == 1 || retry_sequentially) {
```

Where `retry_sequentially` is a `bool` that you set to `true` from the `} catch (newline_error& e) {` branch.

Then it's very clear what the two cases of going through the sequential branch are:

- You have `num_threads = 1`, so no need to waste time on the parallel setup
- You have `retry_sequentially`, where (i think) you expect an error that you surface via the sequential path

src/delimited_index.cc
DavisVaughan: Some small alarm is going off in my head, but it's probably because I don't know this codebase well

How often is `add_parse_error()` elsewhere? Like is there one place where `add_parse_error()` is typically called, and this is a weird one-off ad-hoc calling of `add_parse_error()` just for this special closing quote case?

My question would be about whether we can move the `add_parse_error()` call here closer to whereever it is "normally" called from, but that is likely just wishful thinking?

jennybc: I think you might be banging your shins on this:

> Also: confusingly, we use `add_parse_error()` to record indexing problems. And we use `add_error()` to record problems during parsing. That is neither here not there, but something I keep noticing.)

`add_parse_error()` is *only* used for indexing problems. And every true parsing error you've ever seen from readr/vroom is actually recorded via a different function, `add_error()`! All existing `add_parse_error()` calls live in `delimited_index*.cc/.h` (indexing), while all `add_error()` calls live in `vroom_*.cc/.h` (parsing). The naming is backwards from what you'd expect, but the separation is consistent.

Prior to this PR, the only type of problem that was anticipated to be discovered during indexing was "this record has too few/many fields". I feel OK adjusting `add_parse_error()` to reflect that indexing can surface structural problems other than too few/many fields. If you squint, I guess you could say that an unclosed quote is an important special case of "too few fields".

jennybc: I might try to do some renaming to eliminate this gotcha, in a separate PR.

DavisVaughan: I will be happy with whatever you come up with here, just wanted to raise the metaphorical alarm

PR #624, opened 2026-06-06
Fix problems() support for vroom_lines() output (#607)

### LINKED ISSUE #607: `vroom_lines()` warns about problems but `problems()` doesn't work on its output
When `vroom_lines()` encounters a parsing issue (e.g., an embedded null byte), it emits a warning directing users to call `problems()`. However, `problems()` rejects the character vector returned by `vroom_lines()`, leaving users with no way to get more details on what went wrong.

*Note: This issue is exposed by a fix in #604, which improves null byte handling in `vroom_lines()`. In released vroom, the whole embedded null byte issue is silently swallowed which is also not good.*

``` r
tmp <- withr::local_tempfile()
writeBin(c(charToRaw("line1\n"), as.raw(0x00), charToRaw("line2\n")), tmp)

x <- vroom::vroom_lines(tmp)
x
#> [1]
#> e.g.:
#>   dat <- vroom(...)
#>   problems(dat)
#>  "line1" ""
vroom::problems(x)
#> Error in `vroom::problems()`:
#> ! The `x` argument of `vroom::problems()` must be a data frame created
#>   by vroom:
#> ✖ `x` has class <character>
```

<sup>Created on 2026-01-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

`vroom_lines()` calls `vroom_()` internally, which returns a data frame with a problems attribute. But `vroom_lines()` then extracts just the first column (`out[[1]]`), discarding the problems attribute. Plus problem reporting also refuses to work with a non-data.frame object.

LeonidasZhak:
Fixes #607.

## Problem

When `vroom_lines()` encounters a parsing issue, it warns users to call `problems()`. However:
1. `vroom_lines()` discards the `"problems"` attribute when extracting the character column
2. `problems()` only accepts `tbl_df` objects, rejecting character vectors

## Fix

- `R/vroom_lines.R`: Propagate the `"problems"` attribute from the tibble to the character vector when parsing errors exist
- `R/problems.R`: Accept character vectors with a `"problems"` attribute; improve error messages for unsupported types
- `man/problems.Rd`: Update `@param x` to document character vector support
- `tests/testthat/test-vroom_lines.R`: Add tests for attribute propagation and error messages

## Validation

```
R CMD INSTALL --no-docs --no-multiarch --no-test-load vroom
# vroom_lines tests: [ FAIL 0 | WARN 0 | SKIP 0 | PASS 19 ]
# problems tests: [ FAIL 1 | WARN 0 | SKIP 3 | PASS 51 ] (1 pre-existing failure)
```

The pre-existing failure at `test-problems.R:132` is unrelated (`vroom_materialize` not found in test harness).

### DIFF
--- a/R/problems.R
+++ b/R/problems.R
@@ -5,7 +5,8 @@
 #' might want to know about. You can retrieve a data frame of these problems
 #' with this function.
 #'
-#' @param x A data frame from `vroom::vroom()`.
+#' @param x A data frame from `vroom::vroom()` or a character vector from
+#'   `vroom::vroom_lines()`.
 #' @param lazy If `TRUE`, just the problems found so far are returned. If
 #'   `FALSE` (the default) the lazy data is first read completely and all
 #'   problems are returned.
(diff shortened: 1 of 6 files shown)


LeonidasZhak:
Withdrawing this because CI is currently failing and I should not ask maintainers to review an unready automated PR. I will only come back with a clean, focused change if useful. Sorry for the noise, and thank you for maintaining the project.

PR #625, opened 2026-06-06
fix: allow %e format to parse single-digit days without leading space

### LINKED ISSUE #1559: parse error when using a readr column specification for day
read_delim(I("d1\nJanuary 1, 2010"), delim = "|",
           col_types = cols(d1 = col_date("%B %e, %Y"))) # "%e" not work, but a "%d" works

LeonidasZhak:
## Problem

The `%e` format specifier (day with optional leading space) fails to parse single-digit days without a leading space. For example:

```r
readr::read_csv(I("d1\nJanuary 1, 2010"), col_types = list(d1 = col_date("%B %e, %Y")))
# Returns NA with parsing failure
```

This is because `consumeIntegerWithSpace` calls `consumeInteger(n, pOut)` with the default `exact = true`, which requires consuming exactly `n` characters. A single-digit day like "1" only consumes 1 character out of the expected 2, causing the parse to fail.

Note that `%d` already handles this correctly — it uses `consumeInteger(2, &day_, false)` with `exact = false`.

## Fix

One-line change in `src/DateTimeParser.h`: pass `exact = false` to `consumeInteger` in `consumeIntegerWithSpace`.

## Tests

Added 2 new test cases:
- `%e` parses single-digit day without leading space ("January 1, 2010")
- `%e` parses two-digit day ("January 15, 2010")

Full test suite passes (4 pre-existing failures in test-path.R related to non-ASCII encoding on macOS, unrelated to this change).

## Related

- Fixes tidyverse/readr#1559

### DIFF
--- a/src/DateTimeParser.h
+++ b/src/DateTimeParser.h
@@ -474,7 +474,7 @@ private:
     if (consumeThisChar(' '))
       n--;
 
-    return consumeInteger(n, pOut);
+    return consumeInteger(n, pOut, false);
   }
 
   inline bool consumeDouble(double* pOut) {
--- a/tests/testthat/test-datetime.R
+++ b/tests/testthat/test-datetime.R
@@ -58,6 +58,27 @@ test_that("%e allows leading space", {
   )
 })
 
+test_that("%e parses single-digit day without leading space", {
+  test_parse_date(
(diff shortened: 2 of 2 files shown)


LeonidasZhak:

PR #629, opened 2026-06-06
fix: skip non-ASCII path tests when locale lacks UTF-8 support

LeonidasZhak:
## What

Four tests in `test-path.R` use Unicode characters in `tempfile()` patterns but fail on systems where the locale is `C` (no UTF-8 support), because `tempfile()` cannot translate the patterns to the native encoding.

This PR adds `skip_if(!l10n_info()$`UTF-8`)` guards to these tests:
- can read file w/o final newline, w/ multi-byte characters in path
- can read file w/ final newline, w/ multi-byte characters in path
- can write to path with non-ascii characters
- can read fwf file w/ non-ascii characters in path

Also updates `_snaps/path.md` to match current lifecycle 1.0.5.9000 wording (`in vroom 1.5.0` instead of `as of vroom 1.5.0`).

## Why

On macOS (and other systems) with `LANG=C` or `LC_CTYPE=C`, these tests fail with:

```
unable to translate 'no-trailing-nèwline-mülti-byté-path-' to native encoding
```

The sibling test at line 184 already has a locale guard (`skip_if(l10n_info()$`Latin-1`)`), but the four affected tests had none.

## Validation

```sh
Rscript -e 'devtools::test()'
# FAIL 0 | WARN 0 | SKIP 9 | PASS 1159
```

All 4 previously-failing tests now correctly skip on non-UTF-8 locales.

### DIFF
--- a/tests/testthat/_snaps/path.md
+++ b/tests/testthat/_snaps/path.md
@@ -20,7 +20,7 @@
       x <- vroom("a,b,c,d\n1,2,3,4", show_col_types = FALSE)
     Condition
       Warning:
-      The `file` argument of `vroom()` must use `I()` for literal data as of vroom 1.5.0.
+      The `file` argument of `vroom()` must use `I()` for literal data in vroom 1.5.0.
         
         # Bad:
         vroom("X,Y\n1.5,2.3\n")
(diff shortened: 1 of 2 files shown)


LeonidasZhak:

PR #631, opened 2026-06-07
fix: add [.spec_tbl_df method to drop spec/problems on subsetting (#569)

### LINKED ISSUE #569: Why does vroom not implement `[.spec_tbl_df`?
Observation formed while working on #566 and #567.

readr has a `[.spec_tbl_df` method "To ensure the `spec` is dropped once they [readr-parsed data frames] are subset" (quoting from the readr commit where this was added https://github.com/tidyverse/readr/commit/e94b8bc99c172d7c5b7a4877b3503d3ff01265d5).

Seems to me this would apply also when someone is just using vroom directly, i.e. not via readr. Having such a method requires a bit of care (finesse the usual readr/vroom conditional registration stuff, make sure you don't call `[` internally after you've applied the `"spec_tbl_df"` class, etc.). But it seems like vroom should do this.

readr's `[.spec_tbl_df` method removes the `spec` and `problems` attributes and removes the `"spec_tbl_df"` subclass: 
https://github.com/tidyverse/readr/blob/46438c194f1216d8851eea439b6de7b604c004b7/R/utils.R#L130-L136

LeonidasZhak:
Closes #569.

## What

Add a `[.spec_tbl_df` S3 method to vroom that drops the `spec` and `problems` attributes and removes the `spec_tbl_df` subclass when a data frame is subset with `[`.

## Why

When vroom creates a data frame, it applies the `spec_tbl_df` class and attaches `spec` and `problems` attributes. Without a `[.spec_tbl_df` method, these attributes persist after subsetting, which can confuse users and tools that inspect them. readr already has this method; vroom should too.

## Changes

- `R/utils.R`: Add `[.spec_tbl_df` method (same pattern as `as.data.frame.spec_tbl_df`, `as_tibble.spec_tbl_df`, etc.)
- `R/zzz.R`: Register conditionally via `s3_register("base::[", "spec_tbl_df")` when readr is not loaded
- `tests/testthat/test-vroom.R`: Add test verifying that subsetting drops class and attributes

## Validation

- `devtools::test(filter = "vroom$")`: 354 pass, 0 fail
- `devtools::test()`: 1164 pass, 5 pre-existing failures (locale encoding in test-path.R), 0 new failures
- Verified method works both with and without readr loaded

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -106,6 +106,16 @@ as.data.frame.spec_tbl_df <- function(x, ...) {
   NextMethod("as.data.frame")
 }
 
+# Conditionally exported in zzz.R
+#' @noRd
+# @export
+`[.spec_tbl_df` <- function(x, ...) {
+  attr(x, "spec") <- NULL
+  attr(x, "problems") <- NULL
+  class(x) <- setdiff(class(x), "spec_tbl_df")
+  NextMethod(`[`)
+}
+
 is_rstudio_console <- function() {
   !(Sys.getenv("RSTUDIO", "") == "" || Sys.getenv("RSTUDIO_TERM", "") != "")
 }
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -13,6 +13,7 @@
     s3_register("base::print", "date_names")
     s3_register("base::print", "locale")
     s3_register("utils::str", "col_spec")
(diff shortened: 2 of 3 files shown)


LeonidasZhak:

PR #634, opened 2026-06-07
fix: lone quote in skipped line no longer consumes rest of file

### LINKED ISSUE #1577: read_csv() gives empty table with double quote in skipped line
If I use read_csv() with a skip argument and one of the skipped lines has a lone double quote, the returned tibble is empty. If I delete the quote, or add a second quote, the table is read as expected. read.csv() gives the expected data frame in both cases.

```r
library(readr)
write('"\na,b\n1,2', "quote.csv")
read_csv("quote.csv", skip = 1)
```

LeonidasZhak:
Fixes tidyverse/readr#1577

## Problem

When `skip` is used and a skipped line contains a lone (unclosed) quote character, `find_next_non_quoted_newline` sets `in_quote = true` and never resets it. This causes all subsequent newlines to be treated as embedded within a quoted field, making the skip consume the entire file.

```r
# Before fix: returns empty tibble
write('"\na,b\n1,2', "quote.csv")
read_csv("quote.csv", skip = 1)
#> # A tibble: 0 × 0
```

## Fix

In `find_next_non_quoted_newline` (`src/utils.h`), when EOF is reached while `in_quote` is still true (indicating an unclosed quote), fall back to finding the first newline from the original start position. This ensures a lone quote in a skipped line doesn't consume the rest of the file.

```r
# After fix: returns expected data
write('"\na,b\n1,2', "quote.csv")
read_csv("quote.csv", skip = 1)
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2
```

## Changes

- `src/utils.h`: Add fallback logic in `find_next_non_quoted_newline` for unclosed quotes
- `tests/testthat/test-vroom.R`: Add regression test

## Testing

- New test passes
- All existing tests pass (1163 pass, 5 pre-existing Unicode path encoding failures on macOS)
- Properly quoted fields with embedded newlines still work correctly

### DIFF
--- a/R/vroom_fwf.R
+++ b/R/vroom_fwf.R
@@ -48,6 +48,16 @@
 #' fwf_cols(name = 20, state = 10, ssn = 12)
 #' ```
 #'
+#' @section Multi-byte characters:
+#' Field positions are interpreted as **byte** offsets, not character positions.
+#' For ASCII-only text these are identical, but for multi-byte encodings such as
+#' UTF-8 the two can differ.  For example, the degree symbol `\u00b0` occupies
+#' one character but two bytes.  When a line contains multi-byte characters,
+#' field boundaries specified in character positions will be misaligned.
+#'
+#' If your fixed-width file contains non-ASCII text, count positions by bytes
+#' (e.g. `nchar(x, type = "bytes")`) rather than by characters.
+#'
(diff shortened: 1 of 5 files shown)


LeonidasZhak:

